PR #83, opened 2025-04-15
Mitsubishi Inverter EL6021

### LINKED ISSUE #9: Implement EL6021
- [ ] Device EL6021
  - The device likely needs an internal message queue
- [x]  IO `SerialInterface`
  - May not need `state` and `write` methods but `hasMessage` `readMessage` and `writeMessage`

### LINKED ISSUE #15: Add `MitsubishiInverterRS232` actor

1cedsoda:
fix #9 
fix #15


kraemr:
you can discard the loop.rs changes, i did some tests in there and didnt notice that i committed it, now i would have to rebase

Cargo.toml
```diff
@@ -5,6 +5,7 @@ members = [
     "stepper-driver",
     "ethercat-hal-derive",
     "control-core",
+    "common",
```
1cedsoda: Move code into `control-core`

common/src/modbus/modbus.rs
```diff
@@ -0,0 +1,208 @@
...
+            | Self::Coding8M1 => 1,
+            Self::Coding7E2
+            | Self::Coding7O2
+            | Self::Coding8N2
+            | Self::Coding8E2
+            | Self::Coding8O2 => 2,
+        }
+    }
+
+    /// Get the total number of bits per byte (including start bit)
```
1cedsoda: bits per byte?

common/src/modbus/modbus.rs
```diff
@@ -0,0 +1,208 @@
...
+    Odd,
+    Space,
+    Mark,
+}
+
+fn convert_nanoseconds_to_milliseconds(nanoseconds: u64) -> u64 {
+    if nanoseconds == 0 {
+        return 0;
+    }
+    return nanoseconds / 1000000;
```
1cedsoda: `std::time::Instant` verwenden und dessen conversion Methoden

common/src/modbus/modbus.rs
```diff
@@ -0,0 +1,208 @@
...
+
+pub struct ModbusRequest {
+    pub slave_id: u8,
+    pub function_code: ModbusFunctionCode,
+    pub data: Vec<u8>,
+}
+
+pub struct ModbusResponse {
+    pub slave_id: u8,
+    pub function_code: u8, // needs to be u8 because of exceptions
```
1cedsoda: Kannst auch variante `Unknown(u8)` hinzufügen dann kann `function_code` auch `ModbusFunctionCode` sein.

common/src/modbus/modbus.rs
```diff
@@ -0,0 +1,208 @@
...
+    pub data: Vec<u8>,
+}
+
+impl From<ModbusFunctionCode> for u8 {
+    fn from(value: ModbusFunctionCode) -> Self {
+        match value {
+            ModbusFunctionCode::ReadHoldingRegister => 0x03,
+            ModbusFunctionCode::PresetHoldingRegister => 0x06, // Preset basically just means write
```
1cedsoda: Move comment to `enum ModbusFunctionCode` to describe function

common/src/modbus/modbus.rs
```diff
@@ -0,0 +1,208 @@
...
+    let nanoseconds_per_bit: u64 = (1000000 / baudrate) as u64;
+    let nanoseconds_per_byte: u64 = bits as u64 * nanoseconds_per_bit as u64;
+    let transmission_timeout: u64 = nanoseconds_per_byte * message_size as u64;
+    let silent_time: u64 = (nanoseconds_per_byte * (35)) / 10 as u64; // silent_time is 3.5x of character length,which is 11 bit for 8E1
+    let mut full_timeout: u64 = transmission_timeout as u64;
+    full_timeout += machine_operation_delay_nano;
+    full_timeout += silent_time;
+    return full_timeout as u64;
+}
+
```
1cedsoda: Create test fort this algorithm

common/src/modbus/modbus.rs
```diff
@@ -0,0 +1,208 @@
...
+/// machine_operation_delay_nano: Delay for the given operation in nanoseconds as specified by the slaves datasheet (example: mitsubishi csfr84 has 12ms for read write in RAM)
+/// baudrate: bits per second
+/// message_size: size of original message in bytes
+pub fn calculate_modbus_timeout(
+    bits: u8,
+    machine_operation_delay_nano: u64,
+    baudrate: u32,
+    message_size: u32,
+) -> u64 {
+    let nanoseconds_per_bit: u64 = (1000000 / baudrate) as u64;
```
1cedsoda: use `std::time`

control-core/src/actors/mitsubishi_inverter_rs485.rs
```diff
@@ -0,0 +1,170 @@
...
+
+            if diff < timeout {
+                return;
+            }
+            self.last_ts = now.timestamp_nanos_opt().unwrap();
+
+            if let Operation::Send = self.last_op {
+                let res = (self.serial_interface.read_message)().await;
+
+                println!("read result: {:x?}", res);
```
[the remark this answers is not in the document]
1cedsoda: Yes, use `log` crate (`log::debug!`) or remove print

control-core/src/actors/mitsubishi_inverter_rs485.rs
```diff
@@ -0,0 +1,170 @@
...
+impl Actor for MitsubishiInverterRS485Actor {
+    fn act(&mut self, _now_ts: u64) -> Pin<Box<dyn Future<Output = ()> + Send + '_>> {
+        Box::pin(async move {
+            let now = Utc::now();
+            if self.last_ts == 0 {
+                self.last_ts = now.timestamp_nanos_opt().unwrap();
+            }
+            let diff: i64 = now.timestamp_nanos_opt().unwrap() - self.last_ts;
+
+            // TODO: implement a message queue for the messages coming from machine api AND a response message queue maybe ?
```
1cedsoda: TODO

control-core/src/actors/mitsubishi_inverter_rs485.rs
```diff
@@ -0,0 +1,170 @@
...
+    Receive,
+    None,
+}
+
+#[derive(Debug)]
+pub struct MitsubishiInverterRS485Actor {
+    pub received_response: bool,
+    pub init_done: bool,
+    pub serial_interface: SerialInterface,
+    pub last_ts: i64,
```
1cedsoda: `std::time::Instant`

control-core/src/actors/mitsubishi_inverter_rs485.rs
```diff
@@ -0,0 +1,170 @@
...
+            last_op: Operation::None,
+        }
+    }
+}
+
+pub fn response_is_exception(response: ModbusResponse) -> bool {
+    return (response.function_code & 0b10000000) > 0; // 0x80 is set when an exception happens
+}
+
+pub fn response_functioncode_is_exception(function_code: u8) -> bool {
```
1cedsoda: `function_code` since the `_` is used also elsewhere

control-core/src/actors/mitsubishi_inverter_rs485.rs
```diff
@@ -0,0 +1,170 @@
...
+            1 => MitsubishiModbusExceptionCode::IllegalFunction,
+            2 => MitsubishiModbusExceptionCode::IllegalDataAddress,
+            3 => MitsubishiModbusExceptionCode::IllegalDataValue,
+            _ => MitsubishiModbusExceptionCode::None,
+        }
+    }
+}
+
+impl Actor for MitsubishiInverterRS485Actor {
+    fn act(&mut self, _now_ts: u64) -> Pin<Box<dyn Future<Output = ()> + Send + '_>> {
```
1cedsoda: reduce size of `act` method and move logic to methods

docs/example_communication
1cedsoda, line 1: File ending?

ethercat-hal/src/devices/el6021.rs
```diff
@@ -0,0 +1,470 @@
...
+            7 => Some(EL6021Baudrate::B19200),
+            8 => Some(EL6021Baudrate::B38400),
+            9 => Some(EL6021Baudrate::B57600),
+            10 => Some(EL6021Baudrate::B115200),
+            _ => None,
+        }
+    }
+    pub fn into(self) -> u8 {
+        self as u8
+    }
```
1cedsoda: Use `impl From<...> for u8`

ethercat-hal/src/devices/el6021.rs
```diff
@@ -0,0 +1,470 @@
...
+            4 => Some(EL6021Baudrate::B2400),
+            5 => Some(EL6021Baudrate::B4800),
+            6 => Some(EL6021Baudrate::B9600),
+            7 => Some(EL6021Baudrate::B19200),
+            8 => Some(EL6021Baudrate::B38400),
+            9 => Some(EL6021Baudrate::B57600),
+            10 => Some(EL6021Baudrate::B115200),
+            _ => None,
+        }
+    }
```
1cedsoda: Use `TryFrom` trait

ethercat-hal/src/devices/el6021.rs
```diff
@@ -0,0 +1,470 @@
...
+            | (EL6021Baudrate::B57600, SerialEncoding::Coding7E2)
+            | (EL6021Baudrate::B115200, SerialEncoding::Coding7O2) => {}
+            _ => {
+                return Err(anyhow!(
+                "ERROR: EL6021Configuration::write_config Baudrate and Coding is not compatible!"
+            ))
+            }
+        }
+
+        //   device.sdo_write(0x8000, 0x1, self.rts_enabled).await?;
```
1cedsoda: TODO

ethercat-hal/src/devices/el6021.rs
```diff
@@ -0,0 +1,470 @@
...
+
+impl Default for Standard22ByteMdp600Output {
+    fn default() -> Self {
+        Self {
+            control: 0,
+            length: 0,
+            data: [0u8; 22],
+        }
+    }
+}
```
1cedsoda: Since all values are 0ed you can `#[derive(Default)]`

ethercat-hal/src/devices/el6021.rs
```diff
@@ -0,0 +1,470 @@
...
+            .await?;
+        Ok(())
+    }
+}
+
+/// The value is accompanied by some metadata.
+#[derive(Debug, Clone, PdoObject, PartialEq)]
+#[pdo_object(bits = 192)]
+pub struct Standard22ByteMdp600Output {
+    pub control: u8, // For Standard22ByteMdp600 Output same size as input, but status is ctrl instead
```
1cedsoda: Does the comment make sense?

ethercat-hal/src/devices/el6021.rs
```diff
@@ -0,0 +1,470 @@
...
+            .await?;
+
+        device
+            .sdo_write(0x8000, 0x1b, self.explicit_baudrate)
+            .await?;
+
+        device
+            .sdo_write(0x8000, 0x1c, self.extended_data_frame)
+            .await?;
+        Ok(())
```
1cedsoda, line 281: PDO assigment writing missing

ethercat-hal/src/devices/el6021.rs
```diff
@@ -0,0 +1,470 @@
...
+    }
+}
+
+/// The value is accompanied by some metadata.
```
1cedsoda: ?

ethercat-hal/src/devices/el6021.rs
```diff
@@ -0,0 +1,470 @@
...
+
+impl RxPdoObject for Standard22ByteMdp600Output {
+    fn write(&self, buffer: &mut BitSlice<u8, Lsb0>) {
+        buffer[0..8].store_le(self.control);
+        buffer[8..16].store_le(self.length);
+        for (i, &byte) in self.data.iter().take(22).enumerate() {
+            buffer[(16 + i * 8)..(16 + (i + 1) * 8)].store_le(byte);
+        }
+    }
+}
```
1cedsoda: Add tests

ethercat-hal/src/devices/el6021.rs
1cedsoda: Use `enum` instead of `u8`.

ethercat-hal/src/devices/el6021.rs
```diff
@@ -0,0 +1,470 @@
...
+    pub control: u8, // For Standard22ByteMdp600 Output same size as input, but status is ctrl instead
+    pub length: u8,
+    pub data: [u8; 22],
+}
+
+/// The value is accompanied by some metadata.
+#[derive(Debug, Clone, PdoObject, PartialEq)]
+#[pdo_object(bits = 192)]
+pub struct Standard22ByteMdp600Input {
+    pub status: u8, // For Standard22ByteMdp600 Output same size as input, but status is ctrl instead
```
1cedsoda: `status` to `enum`

ethercat-hal/src/devices/el6021.rs
```diff
@@ -0,0 +1,470 @@
...
+    TransmitAccepted,
+    ReceiveRequest,
+    InitAccepted,
+}
+
+impl SerialInterfaceDevice<EL6021Port> for EL6021 {
+    fn serial_init_request(&mut self, _port: EL6021Port) -> () {
+        if let Some(rx_pdo) = &mut self.rxpdo.com_rx_pdo_map_22_byte {
+            println!("Initializing EL6021 with new settings...");
+            rx_pdo.control |= 0x0004;
```
1cedsoda: Move decoding to enum impl

ethercat-hal/src/devices/el6021.rs
```diff
@@ -0,0 +1,470 @@
...
+            std::thread::sleep(std::time::Duration::from_millis(50));
+            if let Some(tx_pdo) = &self.txpdo.com_tx_pdo_map_22_byte {
+                if (tx_pdo.status & 0x0004) != 0 {
```
1cedsoda: Move decoding to enum impl

ethercat-hal/src/devices/el6021.rs
```diff
@@ -0,0 +1,470 @@
...
+    ReceiveRequest,
+    InitAccepted,
+}
+
+impl SerialInterfaceDevice<EL6021Port> for EL6021 {
+    fn serial_init_request(&mut self, _port: EL6021Port) -> () {
+        if let Some(rx_pdo) = &mut self.rxpdo.com_rx_pdo_map_22_byte {
+            println!("Initializing EL6021 with new settings...");
+            rx_pdo.control |= 0x0004;
+            std::thread::sleep(std::time::Duration::from_millis(50));
```
1cedsoda: Thread sleep is bad.

ethercat-hal/src/devices/el6021.rs
```diff
@@ -0,0 +1,470 @@
...
+            if let Some(tx_pdo) = &self.txpdo.com_tx_pdo_map_22_byte {
+                if (tx_pdo.status & 0x0004) != 0 {
+                    println!("init Accepted {}", tx_pdo.status);
```
1cedsoda: remove prinln

ethercat-hal/src/devices/el6021.rs
```diff
@@ -0,0 +1,470 @@
...
+                    rx_pdo.control = 0;
+                    std::thread::sleep(std::time::Duration::from_millis(50));
```
1cedsoda: Thread sleep bad

ethercat-hal/src/devices/el6021.rs
```diff
@@ -0,0 +1,470 @@
...
+                }
+            }
+        }
+    }
+
+    fn serial_interface_has_messages(&mut self, _port: EL6021Port) -> bool {
+        if let Some(tx_pdo) = &self.txpdo.com_tx_pdo_map_22_byte {
+            return (tx_pdo.status & 0x2) != 0;
```
1cedsoda: Decode elsewhere

ethercat-hal/src/devices/el6021.rs
```diff
@@ -0,0 +1,470 @@
...
+        if !self.serial_interface_has_messages(_port) {
+            println!("has no messages");
+            return vec![];
+        }
+
+        if let Some(tx_pdo) = &mut self.txpdo.com_tx_pdo_map_22_byte {
+            let valid_length = tx_pdo.length as usize;
+            let received_data = tx_pdo.data[..valid_length.min(22)].to_vec();
+            if let Some(rx_pdo) = &mut self.rxpdo.com_rx_pdo_map_22_byte {
+                rx_pdo.control ^= 0x2;
```
1cedsoda: Decode elsehwere

ethercat-hal/src/io/serial_interface.rs
```diff
@@ -0,0 +1,108 @@
...
+use smol::lock::RwLock;
+
+/// Digital Input (DI) device
+///
+/// Reads digital values (true or false) from the device.
+pub struct SerialInterface {
+    pub has_message: Box<dyn Fn() -> Pin<Box<dyn Future<Output = bool> + Send>> + Send + Sync>,
+    pub write_message: Box<dyn Fn( Vec<u8> ) -> Pin<Box<dyn Future<Output = ()> + Send>> + Send + Sync>,
+    pub read_message: Box<dyn Fn() -> Pin<Box<dyn Future<Output = Vec<u8>> + Send>> + Send + Sync>,
+    pub initialize_serial: Box<dyn Fn() -> Pin<Box<dyn Future<Output = ()> + Send>> + Send + Sync>,
```
1cedsoda: Do we really need intialize serial? Think we should remove this method

ethercat-hal/src/io/serial_interface.rs
```diff
@@ -0,0 +1,108 @@
+use std::{fmt, future::Future, pin::Pin, sync::Arc};
+
+use smol::lock::RwLock;
+
+/// Digital Input (DI) device
+///
+/// Reads digital values (true or false) from the device.
+pub struct SerialInterface {
+    pub has_message: Box<dyn Fn() -> Pin<Box<dyn Future<Output = bool> + Send>> + Send + Sync>,
+    pub write_message: Box<dyn Fn( Vec<u8> ) -> Pin<Box<dyn Future<Output = ()> + Send>> + Send + Sync>,
```
1cedsoda: `write_message` should return a `Result` for example if the `Vec` is too long

ethercat-hal/src/io/serial_interface.rs
```diff
@@ -0,0 +1,108 @@
...
+    pub read_message: Box<dyn Fn() -> Pin<Box<dyn Future<Output = Vec<u8>> + Send>> + Send + Sync>,
```
1cedsoda: What happens if `has_message` was ignored in this is called?
Maybe we need to give an `Err` if there is no message or `Option<Vec<u8>>` and omit the `has_message` completely.

ethercat-hal/src/io/serial_interface.rs
```diff
@@ -0,0 +1,108 @@
...
+    pub initialize_serial: Box<dyn Fn() -> Pin<Box<dyn Future<Output = ()> + Send>> + Send + Sync>,
+
+}
+
+impl fmt::Debug for SerialInterface {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "SerialInterface")
+    }
+}
```
1cedsoda, line 24: Use derive

ethercat-hal/src/pdo/mod.rs
```diff
@@ -2,7 +2,7 @@ pub mod basic;
 pub mod el252x;
 pub mod el30xx;
 pub mod el32xx;
-pub mod el7031;
+//pub mod el7031;
```
1cedsoda: ???

server/src/ethercat/loop.rs
```diff
@@ -99,11 +107,29 @@ pub async fn setup_loop(interface: &str, app_state: Arc<AppState>) -> Result<(),
         );
     }
 
+    println!("{:?}",&subdevices);
```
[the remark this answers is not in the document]
1cedsoda: What he sais

server/src/ethercat/loop.rs
1cedsoda, line 1: Remove all changes from this file

common/src/modbus/modbus.rs
1cedsoda, line 1: I just see encoding, where is the decoding of modbus responses (`impl TryFrom<Vec<u8>> for ModbusResponse`)

1cedsoda:
Please merge `master` into your branch to resolve conflicts.
There have been some important changes like removal of `input_ts` and `output_ts`

1cedsoda:
Mainly, remove non-production code and code quality improvements. Also remove the `common` crate and move the code to `control-core`

common/src/modbus/modbus.rs
kraemr: When transferring 1 byte in 8N1 Encoding it would be 10 bits that get transferred over serial (8 data bits + 1 start bit + 1 stop bit). Could probably be a little more specific

Cargo.toml
kraemr: sadly just moving the code into control-core creates a cyclical dependency

docs/example_communication
kraemr: this was a communication of PC directly with inverter. This is a pulseview session file, can be removed

common/src/modbus/modbus.rs
kraemr: 6b208f5197286191f43cfb80e251f4ffdaa1f1f0

common/src/modbus/modbus.rs
kraemr: 73cac1c177332dd5564ef38399c486b801f9e738

docs/example_communication
kraemr: e442d66d176150be675fdb79072b2df0f4a1c1a2

common/src/modbus/modbus.rs
kraemr: cbc135433ae9bf1f967b2f852aa8e3de93caa383

common/src/modbus/modbus.rs
kraemr: cbc135433ae9bf1f967b2f852aa8e3de93caa383

ethercat-hal/src/devices/el6021.rs
kraemr: 4d59337eef4a73574772191f9251324bdc43bcd0

common/src/modbus/modbus.rs
kraemr: 991416d557ab911d56b8dc863b01211aada1dd6a

control-core/src/actors/mitsubishi_inverter_rs485.rs
kraemr: 991416d557ab911d56b8dc863b01211aada1dd6a

control-core/src/actors/mitsubishi_inverter_rs485.rs
kraemr: 9083cc6af66d69e3bd23ffe07235784a0862b181

ethercat-hal/src/devices/el6021.rs
kraemr: dbb2e506959bcd647c365a5386cfd51181367185

ethercat-hal/src/devices/el6021.rs
kraemr: 59537c0ea3b462558b7d78218dd7e5449b6d289d

ethercat-hal/src/devices/el6021.rs
kraemr: 897c32d3162531cf9df8d0aa016f841b7a4042fb

ethercat-hal/src/devices/el6021.rs
kraemr: 897c32d3162531cf9df8d0aa016f841b7a4042fb

ethercat-hal/src/io/serial_interface.rs
kraemr: Yes we should remove it for now. The El6021 has the possibility of changing its baudrate, parity etc while its running by setting the init bit in control.

kraemr: 5ca501b224de39ba17fb303474c61c339ac0c182

ethercat-hal/src/devices/el6021.rs
kraemr: 5ca501b224de39ba17fb303474c61c339ac0c182

ethercat-hal/src/devices/el6021.rs
kraemr: awake thread good 5ca501b224de39ba17fb303474c61c339ac0c182

ethercat-hal/src/devices/el6021.rs
kraemr: yep 5ca501b224de39ba17fb303474c61c339ac0c182

ethercat-hal/src/io/serial_interface.rs
kraemr: I think Option<Vec<u8>> is the best fit, as no message read doesnt necessarily mean that there is an error, because there are also modbus requests were no response happens

ethercat-hal/src/devices/el6021.rs
kraemr: 5ca501b224de39ba17fb303474c61c339ac0c182

ethercat-hal/src/devices/el6021.rs
kraemr: https://github.com/qitechgmbh/control/commit/5ca501b224de39ba17fb303474c61c339ac0c182

ethercat-hal/src/io/serial_interface.rs
kraemr: this what you had in mind?: 10910f2b66ec3f732660577f87db4b9906258556

ethercat-hal/src/io/serial_interface.rs
kraemr: 10910f2b66ec3f732660577f87db4b9906258556

ethercat-hal/src/devices/el6021.rs
kraemr: 1a616eff74b585d7e00ae0cbf424a0e84c091c6e

ethercat-hal/src/devices/el6021.rs
kraemr: enum doesnt fit there, because we need to be able to set specific bits. We could rewrite it to be multiple bools instead

ethercat-hal/src/devices/el6021.rs
kraemr: enum doesnt fit there, because we need to be able to set specific bits. We could rewrite it to be multiple bools instead

ethercat-hal/src/devices/el6021.rs
kraemr: decode what elsewhere ?

control-core/src/actors/mitsubishi_inverter_rs485.rs
```diff
@@ -0,0 +1,171 @@
...
+                19200,
+                start_motor.len() as u32,
+            );
+
+            // rust-analyzer shows error but it compiles?
+            if elapsed.as_nanos() < timeout.as_nanos() {
+                return;
+            }
+
+            self.last_ts = Instant::now();
```
1cedsoda: use `act` method param `now_ts`

control-core/src/actors/mitsubishi_inverter_rs485.rs
```diff
@@ -0,0 +1,171 @@
...
+                SerialEncoding::Coding8E1.total_bits(),
+                request_timeout,
+                19200,
+                start_motor.len() as u32,
+            );
+
+            // rust-analyzer shows error but it compiles?
+            if elapsed.as_nanos() < timeout.as_nanos() {
+                return;
+            }
```
1cedsoda: I think converting to nanos is not needed.

control-core/src/actors/mitsubishi_inverter_rs485.rs
```diff
@@ -0,0 +1,171 @@
...
+            .into();
+
+            // Maybe encapsulate ModbusRequest inside of MitsubishiRequest, since the timeouts and operations are specific to mitsubishi
+            let request_timeout = RequestType::OperationCommand.timeout_nanoseconds();
+
+            let timeout = calculate_modbus_timeout(
+                SerialEncoding::Coding8E1.total_bits(),
+                request_timeout,
+                19200,
```
1cedsoda: what is this constant

control-core/src/actors/mitsubishi_inverter_rs485.rs
```diff
@@ -0,0 +1,171 @@
...
+impl Actor for MitsubishiInverterRS485Actor {
+    fn act(&mut self, _now_ts: Instant) -> Pin<Box<dyn Future<Output = ()> + Send + '_>> {
+        Box::pin(async move {
+            let elapsed: Duration = self.last_ts.elapsed();
+
+            // TODO: implement a message queue for the messages coming from machine api AND a response message queue maybe ?
+            let start_motor: Vec<u8> = ModbusRequest {
+                slave_id: 1,
+                function_code: ModbusFunctionCode::PresetHoldingRegister,
+                data: vec![0x0, 0x8, 00, 02],
```
1cedsoda: Make an enum with `MitsubishiInverterRS48Requests` which has a variant `StartMotor` (or something else), which  implements `From<MitsubishiInverterRS48Requests> for ModbusRequest`. The `vec!` also is a magic constant that needs to be explained.

common/src/modbus/modbus.rs
kraemr: 36afd1c2e34531790d22f44d92ae5b25ae2b99d9

control-core/src/actors/mitsubishi_inverter_rs485.rs
```diff
@@ -0,0 +1,171 @@
...
+
+            // rust-analyzer shows error but it compiles?
+            if elapsed.as_nanos() < timeout.as_nanos() {
+                return;
+            }
+
+            self.last_ts = Instant::now();
+
+            if let Operation::Send = self.last_op {
+                let res = (self.serial_interface.read_message)().await;
```
1cedsoda: use a `TryFrom` implementatino to try to convert the `Vec<u8>` to a `ModbusResponse`. This should also eliminate the `response_function_code_is_exception` and give a `Result<ModbusResponse, MitsubishiModbusExceptionCode>`.

control-core/src/actors/mitsubishi_inverter_rs485.rs
```diff
@@ -0,0 +1,171 @@
...
+
+                if let Some(result) = res {
+                    if result.len() > 2 {
+                        let is_exception = response_function_code_is_exception(result[1]);
+                        let exception_code = MitsubishiModbusExceptionCode::from(result[2]);
+                        if is_exception {
+                            println!("Mitsubishi Modbus Exception: {}", exception_code.display());
```
1cedsoda: remove `println!`

control-core/src/actors/mitsubishi_inverter_rs485.rs
```diff
@@ -0,0 +1,171 @@
...
+                        }
+                    }
+                }
+
+                // self.last_op = Operation::Receive;
+            }
+            if self.init_done == false {
+                println!("write");
```
1cedsoda: remove `println!`

control-core/src/actors/mitsubishi_inverter_rs485.rs
```diff
@@ -0,0 +1,171 @@
...
+                        let exception_code = MitsubishiModbusExceptionCode::from(result[2]);
+                        if is_exception {
+                            println!("Mitsubishi Modbus Exception: {}", exception_code.display());
+                        }
+                    }
+                }
+
+                // self.last_op = Operation::Receive;
+            }
+            if self.init_done == false {
```
1cedsoda: Maybe instead of `init_done` this struct should have a field `state: Mistsubishi....State` which is an enum of the possible states like `::Uninitalized`, `::Ready` or whatever. This could also solve the problem of the thread sleep usage you implemented by adding an additional state that maks waiting for a cooldown and holds a timetampt of some kind.

So the `act` function will be a big `match` statement

common/src/modbus/modbus.rs
```diff
@@ -0,0 +1,218 @@
...
+        let stop = self.stop_bits();
+
+        format!(
+            "{}{}{}({} bits total)",
+            data,
+            parity,
+            stop,
+            self.total_bits()
+        )
+    }
```
1cedsoda: Seems like dead code? If you need this you could add a `Display` implementation instead.

common/src/modbus/modbus.rs
```diff
@@ -0,0 +1,218 @@
...
+
+fn convert_nanoseconds_to_milliseconds(nanoseconds: u64) -> u128 {
+    let duration: Duration = Duration::from_nanos(nanoseconds);
+    return duration.as_millis();
+}
+
+fn convert_milliseconds_to_nanoseconds(milliseconds: u64) -> u128 {
+    let duration: Duration = Duration::from_millis(milliseconds);
+    return duration.as_nanos();
+}
```
1cedsoda: Dead code.

common/src/modbus/modbus.rs
```diff
@@ -0,0 +1,218 @@
...
+    DiagnoseFunction,      // The response should echo back your request
+    Unknown(u8),           //possibly an exception or a user defined Function
+}
+
+pub struct ModbusRequest {
+    pub slave_id: u8,
+    pub function_code: ModbusFunctionCode,
+    pub data: Vec<u8>,
+}
+// could just make a ModbusFrame instead?
```
1cedsoda: ?

common/src/modbus/modbus.rs
```diff
@@ -0,0 +1,218 @@
...
+    let duration: Duration = Duration::from_millis(milliseconds);
+    return duration.as_nanos();
+}
+
+pub enum ModbusFunctionCode {
+    ReadHoldingRegister,   // Read one or more Registers
+    PresetHoldingRegister, // write one Register Value
+    DiagnoseFunction,      // The response should echo back your request
+    Unknown(u8),           //possibly an exception or a user defined Function
+}
```
1cedsoda: Turn comments into `///` doc comments

common/src/modbus/modbus.rs
```diff
@@ -0,0 +1,218 @@
...
+
+pub struct ModbusRequest {
+    pub slave_id: u8,
+    pub function_code: ModbusFunctionCode,
+    pub data: Vec<u8>,
+}
+// could just make a ModbusFrame instead?
+pub struct ModbusResponse {
+    pub slave_id: u8,
+    pub function_code: ModbusFunctionCode, // needs to be u8 because of exceptions
```
1cedsoda: remove comment

common/src/modbus/modbus.rs
```diff
@@ -0,0 +1,218 @@
...
+    }
+}
+
+/// bits: amount of bits sent for a 8n1 coding: 8 data bits, 0 parity, 1 stop bit (1 start,1 stop) -> 10 bits
+/// machine_operation_delay_nano: Delay for the given operation in nanoseconds as specified by the slaves datasheet (example: mitsubishi csfr84 has 12ms for read write in RAM)
+/// baudrate: bits per second
+/// message_size: size of original message in bytes
+pub fn calculate_modbus_timeout(
+    bits: u8,
+    machine_operation_delay_nano: u64,
```
1cedsoda: Use `Duration` instead of `u64`

common/src/modbus/modbus.rs
```diff
@@ -0,0 +1,218 @@
...
+        buffer.push(high_byte);
+        return buffer;
+    }
+}
+
+/// bits: amount of bits sent for a 8n1 coding: 8 data bits, 0 parity, 1 stop bit (1 start,1 stop) -> 10 bits
+/// machine_operation_delay_nano: Delay for the given operation in nanoseconds as specified by the slaves datasheet (example: mitsubishi csfr84 has 12ms for read write in RAM)
+/// baudrate: bits per second
+/// message_size: size of original message in bytes
+pub fn calculate_modbus_timeout(
```
1cedsoda: a comment for me would be interesting why we need to _calculate_ a timeout.

1cedsoda:
image

common/src/modbus/modbus.rs
```diff
@@ -0,0 +1,274 @@
...
+    bits: u8,
+    machine_operation_delay_nano: u64,
+    baudrate: u32,
+    message_size: u32,
```
1cedsoda: use `message_size: usize` in parameter

common/src/modbus/modbus.rs
```diff
@@ -0,0 +1,274 @@
...
+        buffer.extend_from_slice(&request.data);
+        let length = buffer.len();
+        let result = crc16_modbus.checksum(&buffer[..length]);
+        let high_byte = (result >> 8) as u8; // upper 8 bits
+        let low_byte = (result & 0xFF) as u8; // lower 8 bits
+        buffer.push(low_byte);
+        buffer.push(high_byte);
+        return buffer;
+    }
+}
```
1cedsoda: Tidy up this function a bit. Make space between sections and add comments to explain what is going on. Also move `let crc16_modbus` over `let result` to show the connection.

common/src/modbus/modbus.rs
```diff
@@ -0,0 +1,274 @@
...
+    fn from(request: ModbusRequest) -> Self {
+        let mut buffer = Vec::new();
+        let crc16_modbus: crc::Crc<u16> = crc::Crc::<u16>::new(&crc::CRC_16_MODBUS);
+        buffer.push(request.slave_id);
+        buffer.push(request.function_code.into());
+        buffer.extend_from_slice(&request.data);
+        let length = buffer.len();
+        let result = crc16_modbus.checksum(&buffer[..length]);
+        let high_byte = (result >> 8) as u8; // upper 8 bits
+        let low_byte = (result & 0xFF) as u8; // lower 8 bits
```
1cedsoda: use `to_le_bytes` and then `extend_from_slice`

1cedsoda:
image

Cargo.toml
1cedsoda: Then you put some code in the wrong place.

Move `SerialEncoding` from `common` to `ethercat-hal`. It doesn't even have anything todo with modbus. Then you can remove the dependency from `etehrcat-hal`, move the code to `control-core` and eliminate `common`.

ethercat-hal/src/devices/el6021.rs
```diff
@@ -0,0 +1,470 @@
...
+            5 => Ok(EL6021Baudrate::B4800),
+            6 => Ok(EL6021Baudrate::B9600),
+            7 => Ok(EL6021Baudrate::B19200),
+            8 => Ok(EL6021Baudrate::B38400),
+            9 => Ok(EL6021Baudrate::B57600),
+            10 => Ok(EL6021Baudrate::B115200),
+            _ => Err(()),
+        }
+    }
+}
```
1cedsoda: use `anyhow!` to return a proper error not just `()`

common/src/modbus/modbus.rs
```diff
@@ -0,0 +1,274 @@
...
+impl From<ModbusFunctionCode> for u8 {
+    fn from(value: ModbusFunctionCode) -> Self {
+        match value {
+            ModbusFunctionCode::ReadHoldingRegister => 0x03,
+            ModbusFunctionCode::PresetHoldingRegister => 0x06,
+            ModbusFunctionCode::DiagnoseFunction => 0x08,
+            ModbusFunctionCode::Unknown(value) => value,
+        }
+    }
+}
```
1cedsoda: Sorry, I previously said otherwise but I notices that `Unknown` is an error state that should not be between valid state. So it would be a great idea to turn this into a `TryFrom` also and return an `anyhow!` error for this case.

common/src/modbus/modbus.rs
kraemr: here it is: 3e54671e2a082a725de9214e27a1ccde3bfbd80c

kraemr: Also added some tests

common/src/modbus/modbus.rs
kraemr: Alright no problem

ethercat-hal/src/devices/el6021.rs
kraemr: yup my bad 4699e494cf9439834afd5ed036b0e9262fa34e9e

common/src/modbus/modbus.rs
kraemr: 38b949e187efbed625298e4de332810abfa62113

common/src/modbus/modbus.rs
kraemr: 38b949e187efbed625298e4de332810abfa62113

common/src/modbus/modbus.rs
kraemr: 4625e293953dd616672a32ddf4174d8fda50d8ed

common/src/modbus/modbus.rs
kraemr: Here you go: 1a04bf49e8b1f8d93c418df53a9b2ef31666db54

common/src/modbus/modbus.rs
kraemr: 068a7c3c01cd56231bbdd83857ee9bbb710918ec

control-core/src/actors/mitsubishi_inverter_rs485.rs
kraemr: cff23063b8bcbe75a5879777cf105482ffa5519b

control-core/src/actors/mitsubishi_inverter_rs485.rs
kraemr: yep it is not: 6a256e1e6eebba4c82ed097d8c3678d95921628a

ethercat-hal/src/devices/el6021.rs
```diff
@@ -0,0 +1,472 @@
...
+    /// default: `false`
+    pub xon_on_supported_tx: bool,
+
+    /// # 0x8000:04 - Allows continuous transmission of data from the FIFO buffer.
+    /// - `true` = FIFO continuous send enabled
+    /// - `false` = FIFO continuous send disabled
+    /// default: `false`
+    pub fifo_continuous_send_enabled: bool,
+
+    pub enable_transfer_rate_optimization: bool,
```
1cedsoda, line 167: missing docs

ethercat-hal/src/devices/el6021.rs
```diff
@@ -0,0 +1,472 @@
...
+    /// - `true` = Point-to-point connection mode enabled
+    /// - `false` = Point-to-point connection mode disabled
+    /// default: `false`
+    pub point_to_point_connection_enabled: bool,
+
+    /// # 0x8000:11 - Sets the baud rate (e.g., 9600, 115200).
+    /// This value is typically an index referencing predefined baud rates.
+
+    /// default: `0x06`
+    pub baud_rate: EL6021Baudrate,
```
1cedsoda, line 184: some missing docs

ethercat-hal/src/devices/el6021.rs
```diff
@@ -0,0 +1,472 @@
...
+    pub data_frame: SerialEncoding,
+
+    /// # 0x8000:1A - Notification threshold for the RX buffer (in bytes).
+    /// Determines when the terminal signals the controller that the receive buffer is full.
+    /// default: `0x0360`
+    pub rx_buffer_full_notification: u16,
+
+    /// # 0x8000:1B - Explicitly sets a custom baud rate (only supported from firmware version FW09).
+    /// default: `9600` 0x00000384 0x00002580
+    pub explicit_baudrate: u32,
```
1cedsoda: Does this override the `baud_rate` to a custom u32? If so we need additional logic at some palces. After all a method `get_baud_rate` on this struct that return a `u32` and known when to use the `baur_rate` and when to use `explicit_baudrate`.

ethercat-hal/src/devices/el6021.rs
```diff
@@ -0,0 +1,472 @@
...
+
+    /// # 8000:1C**
```
1cedsoda: remove `**`?

ethercat-hal/src/devices/el6021.rs
```diff
@@ -0,0 +1,472 @@
...
+    /// In this object special formats can also be selected in addition
+    /// to the usual data frames (e.g. 9N1). Changes to this object
+    /// are also adopted in the objects 0x8000:15 and 0x4074.
+    pub extended_data_frame: u16,
```
1cedsoda: Should this be an enum?

ethercat-hal/src/devices/el6021.rs
```diff
@@ -0,0 +1,472 @@
...
+        // baud rate and data frame are only 4 BITs maybe this causes a problem?
+        let baudrate = match EL6021Baudrate::try_from(self.baud_rate) {
+            Ok(baudrate) => baudrate,
+            Err(_) => {
+                println!("Invalid baudrate value: {:?}", self.baud_rate);
+                EL6021Baudrate::B9600 // Default value
+            }
+        };
+        let baudrate_coe_value = u8::from(baudrate);
+
```
1cedsoda: Why converting `baud_rate` from `EL6021Baudrate` to `EL6021Baudrate`?

ethercat-hal/src/devices/mod.rs
```diff
@@ -57,12 +59,22 @@ pub trait Device: NewDevice + Any + Send + Sync + Debug {
...
+    /// Devices can override this function if they want to post processs the input data
+    /// This might be the case if the pdo is not what is needed in the io layer
+    fn input_post_process(&mut self) -> Result<(), anyhow::Error> {
         Ok(())
     }
 
+    /// Devices can override this function if they want to pre processs the output data
+    /// This might be the case if the pdo is not what is needed in the io layer
+    fn output_pre_process(&mut self) -> Result<(), anyhow::Error> {
+        return Ok(());
```
1cedsoda: Use same return style as `input_post_process`. I don't care about using `return` keyword or not but codebase should be uniform.

ethercat-hal/src/io/serial_interface.rs
```diff
@@ -0,0 +1,93 @@
+use anyhow::Error;
+use smol::lock::RwLock;
+use std::{fmt, future::Future, pin::Pin, sync::Arc};
+
+pub struct SerialInterface {
+    pub has_message: Box<dyn Fn() -> Pin<Box<dyn Future<Output = bool> + Send>> + Send + Sync>,
```
1cedsoda, line 6: Can we remove `has_message` or do we need it since `read_message` is an `Option<Vec<u8>>`? Does `read_message` pop a queue and `has_message` does not?

server/src/ethercat/loop.rs
1cedsoda: k

ethercat-hal/src/devices/el6021.rs
1cedsoda: `status` seems to be bit flags. Binary options directly on integers is bad. We first want to decode the bitflags into an enum or struct. Then we can use `if` or `match`.

control-core/src/actors/mitsubishi_inverter_rs485.rs
```diff
@@ -0,0 +1,167 @@
...
+            3 => MitsubishiModbusExceptionCode::IllegalDataValue,
+            _ => MitsubishiModbusExceptionCode::None,
+        }
+    }
+}
+
+impl Actor for MitsubishiInverterRS485Actor {
+    fn act(&mut self, _now_ts: Instant) -> Pin<Box<dyn Future<Output = ()> + Send + '_>> {
+        Box::pin(async move {
+            let elapsed: Duration = self.last_ts.elapsed();
```
1cedsoda: use `duration_since` with `now_ts` because elapsed is internally creating a new timestamp. It's not critical or _bad_ just inconsistent with later setting `last_ts` to `now_ts`.

control-core/src/actors/mitsubishi_inverter_rs485.rs
```diff
@@ -0,0 +1,167 @@
...
+            1 => MitsubishiModbusExceptionCode::IllegalFunction,
+            2 => MitsubishiModbusExceptionCode::IllegalDataAddress,
+            3 => MitsubishiModbusExceptionCode::IllegalDataValue,
+            _ => MitsubishiModbusExceptionCode::None,
+        }
+    }
+}
+
+impl Actor for MitsubishiInverterRS485Actor {
+    fn act(&mut self, _now_ts: Instant) -> Pin<Box<dyn Future<Output = ()> + Send + '_>> {
```
1cedsoda: rename `_now_ts` to `now_ts`. The underscore prefix is used for unused variables so the compiler/clippy doesn't complain with a warning when a var/param is not used. Since we now use the param we can omit the `_`.

control-core/src/actors/mitsubishi_inverter_rs485.rs
kraemr: a63247348eb92872049d99eb051584874328caca

control-core/src/actors/mitsubishi_inverter_rs485.rs
kraemr: it was the baudrate, now serial_interface has serial_get_baudrate and serial_get_coding a63247348eb92872049d99eb051584874328caca

control-core/src/actors/mitsubishi_inverter_rs485.rs
kraemr: 3e54671e2a082a725de9214e27a1ccde3bfbd80c

common/src/modbus/modbus.rs
kraemr: b2b242e9ed114260b54790db66209ec648164ad0

common/src/modbus/modbus.rs
kraemr: 609560b591d33d031b3214d8150cfb0d4b24f37a

control-core/src/actors/mitsubishi_inverter_rs485.rs
kraemr: fd63e9a749e91ce76c1f58491a7c05e9c9c38a97

ethercat-hal/src/devices/el6021.rs
kraemr: 625f37a713e3d25302459ef723b0bdd0ed2dc26f

ethercat-hal/src/devices/el6021.rs
kraemr: 625f37a713e3d25302459ef723b0bdd0ed2dc26f

ethercat-hal/src/devices/mod.rs
kraemr: my bad 625f37a713e3d25302459ef723b0bdd0ed2dc26f

ethercat-hal/src/devices/el6021.rs
kraemr: drunk

ethercat-hal/src/devices/el6021.rs
kraemr: 625f37a713e3d25302459ef723b0bdd0ed2dc26f

control-core/src/actors/mitsubishi_inverter_rs485.rs
kraemr: 0af50d66de770cd675374823eb76c1f0185b9a79

kraemr: 9d61d991b247741012922552e843ca64ba691fea

control-core/src/actors/mitsubishi_inverter_rs485.rs
```diff
@@ -0,0 +1,327 @@
...
+            MitsubishiSystemRegister::AllParameterClear => 0x3,
+            MitsubishiSystemRegister::ParamClearNonCommunication => 0x5,
+            MitsubishiSystemRegister::AllParameterClearNonCommunication => 0x6,
+            MitsubishiSystemRegister::InverterStatusAndControl => 0x8,
+            MitsubishiSystemRegister::OperationModeAndSetting => 0x9,
+            MitsubishiSystemRegister::RunningFrequencyRAM => 0x0d,
+            MitsubishiSystemRegister::RunningFrequencyEEPROM => 0x0e,
+        }
+    }
+}
```
1cedsoda: Yes enum here is okay but it seems like it's analogous to `MitsubishiControlRequests`. In this case I would add a fucntion like `get_system_register` on `MitsubishiControlRequests` if it maps 1:1.

control-core/src/actors/mitsubishi_inverter_rs485.rs
```diff
@@ -0,0 +1,327 @@
...
+
+// TODO: Clean this up
```
1cedsoda: TODO

control-core/src/actors/mitsubishi_inverter_rs485.rs
```diff
@@ -0,0 +1,327 @@
...
+}
+
+#[derive(Debug)]
+pub struct MitsubishiInverterRS485Actor {
+    pub serial_interface: SerialInterface,
+    pub last_ts: Instant,
+    pub last_message_size: usize,
+    pub state: State,
+
+    // do we need an Async Queue? Like a smol::channel or something like that ?
```
1cedsoda: `smol::channel` is for cross thread/task communication. So `VecDequeue` is the right thing to use. Pls remove the comment

control-core/src/actors/mitsubishi_inverter_rs485.rs
```diff
@@ -0,0 +1,327 @@
...
+        }
+    }
+
+    /// This would get called by the api to add a new request to the inverter
+    pub fn add_request(&mut self, request: ModbusRequest) {
+        self.request_queue.push_front(request);
+    }
+
+    /// This is used by the Api to pop off the Response of our Request
+    /// Perhaps we need a transactionId to make sure that the response we got is the correct one ?
```
1cedsoda: Since the send and the recieve would happen in different cycles. We would need to track the transactionId somehow. Since it's a 1-to-1 communication I don't think tracking individual requests have any benefits in this case. So its good as it is. Pls remove the comment

control-core/src/actors/mitsubishi_inverter_rs485.rs
```diff
@@ -0,0 +1,327 @@
...
+// Cant do constant structs in rust with allocations ...
+impl MitsubishiControlRequests {
+    fn get(&self) -> ModbusRequest {
+        match self {
+            MitsubishiControlRequests::ResetInverter => {
+                let reg: u16 = MitsubishiSystemRegister::InverterReset.into();
+                ModbusRequest {
+                    slave_id: 1,
+                    function_code: ModbusFunctionCode::PresetHoldingRegister,
+                    data: vec![reg.to_be_bytes()[0], reg.to_be_bytes()[1], 0x00, 0x01], // Any Value
```
1cedsoda: remove the double call of `to_be_bytes`

control-core/src/actors/mitsubishi_inverter_rs485.rs
```diff
@@ -0,0 +1,327 @@
...
+            request_queue: VecDeque::new(),
+            response_queue: VecDeque::new(),
+            last_message_size: 0,
+        }
+    }
+
+    /// This would get called by the api to add a new request to the inverter
+    pub fn add_request(&mut self, request: ModbusRequest) {
+        self.request_queue.push_front(request);
+    }
```
1cedsoda, line 192: This interface is needed and pretty raw. I think in addition we should add a function `set_frequency` and `get_frequency` which take a `f64` and abstract away the stopping, starting in forward rotation and starting in backward rotation. So we can funnel a frequency right in.

I think a straight-forward function call for all requests is good practivce here like `reset()`, `clear...`get_status` and so on.

The getter functions present a special challenge here since we can't request and respond in the same cycle. We will have to think about solutions here, it's not that trivial.

Also, what happens when per cycle we add two requests but we can only send one? This would fill up the request queue forever. One thing we can do about it is only allowing one request per type. Thinking about this maybe we should convert the `VecDequeue` to a `HashMap`...? Also the response queue since this would make it easier to read specific requests without iterating the whole queue. Also we should maybe implement some logic to give priority to some requests like stop requests since this could pose a physical or health risk to the machine and the operators.

If `set_speed` takes a float we cant `== 0.0` so wee need a range which we interprete as the stop range. Maybe there is a minimum speed in the datasheet, maybe we make it configurable or maybe we make it `F64::EPSILON`. Negative speed would obviously reverse the direction. I would say we give a -1 to 1 to this function and let the user configure what 1 means with a configurationa value which is Hertz from the `uom` crate. We should also implement logic which imposes a maximum acceleration/deceleration with the unit 1/s. Maybe we should add a param `instant_stop: bool` which ignores max deceleration when a speed of `0.0` is applied.

Sorry that's a lot to take in but the interface should be simple to use.
The conversion stuff is similar to the analog input terminals and the speed control is implemented (but distributed and with a PID` on the winder spool motor.

control-core/src/actors/mitsubishi_inverter_rs485.rs
```diff
@@ -0,0 +1,327 @@
...
+        Box::pin(async move {
+            let res: Result<ModbusResponse, _> = (self.serial_interface.read_message)()
+                .await
+                .unwrap()
+                .try_into();
+
+            match res {
+                Ok(result) => {
+                    self.response_queue.push_front(result.clone());
+                    // TODO: make conversion function for ModbusResponse to Vec<u8> or make a len() or size() function or something
```
1cedsoda: why

control-core/src/actors/mitsubishi_inverter_rs485.rs
```diff
@@ -0,0 +1,327 @@
...
+        }
+    }
+}
+
+impl Actor for MitsubishiInverterRS485Actor {
+    fn act(&mut self, now_ts: Instant) -> Pin<Box<dyn Future<Output = ()> + Send + '_>> {
+        Box::pin(async move {
+            let elapsed: Duration = self.last_ts.elapsed();
+            let baudrate = (self.serial_interface.get_baudrate)().await.unwrap();
+            let coding = (self.serial_interface.get_serial_coding)().await.unwrap();
```
1cedsoda: maybe we can save this inside an `Option` and do it only once.

control-core/src/actors/mitsubishi_inverter_rs485.rs
```diff
@@ -0,0 +1,327 @@
...
+                return;
+            }
+            self.last_ts = now_ts;
+            if let State::WaitingForResponse = self.state {
+                self.read_modbus_response().await;
+            }
+
+            if let State::ReadyToSend = self.state {
+                self.send_modbus_request().await;
+            }
```
1cedsoda: make it a `match` statement

ethercat-hal/src/devices/mod.rs
```diff
@@ -57,9 +59,19 @@ pub trait Device: NewDevice + Any + Send + Sync + Debug {
                 input_len
             ));
         }
+        self.input(input);
+        Ok(())
+    }
 
-        self.input(input)?;
```
1cedsoda: Isn't the input `Result` ignored now? I think it was correct beforehand.

ethercat-hal/src/io/serial_interface.rs
```diff
@@ -0,0 +1,134 @@
...
+        dyn Fn(Vec<u8>) -> Pin<Box<dyn Future<Output = Result<(), Error>> + Send>> + Send + Sync,
+    >,
+    pub read_message:
+        Box<dyn Fn() -> Pin<Box<dyn Future<Output = Option<Vec<u8>>> + Send>> + Send + Sync>,
+
+    pub get_baudrate:
+        Box<dyn Fn() -> Pin<Box<dyn Future<Output = Option<u32>> + Send>> + Send + Sync>,
+
+    pub get_serial_coding:
+        Box<dyn Fn() -> Pin<Box<dyn Future<Output = Option<SerialEncoding>> + Send>> + Send + Sync>,
```
1cedsoda: If the return type is named encoding the function shoud also named something with `encoding` not `coding`.

ethercat-hal/src/devices/el6021.rs
```diff
@@ -0,0 +1,499 @@
...
+    /// Legacy 5 Byte MDP 600
+    Legacy5ByteMdp600,
+    /// Standard 22 Byte MDP 600
+    Standard22ByteMdp600,
+    /// Standard 50 Word MDP 600
+    Standard50WordMdp600,
+    /// Standard 98 Byte MDP 600
+    Standard98ByteMdp600,
+    /// No assignment
+    None,
```
1cedsoda: There can't be a `None` assignment

ethercat-hal/src/devices/el6021.rs
```diff
@@ -0,0 +1,499 @@
...
+            EL6021Baudrate::B2400 => 2400,
+            EL6021Baudrate::B4800 => 4800,
+            EL6021Baudrate::B9600 => 9600,
+            EL6021Baudrate::B19200 => 19200,
+            EL6021Baudrate::B38400 => 38400,
+            EL6021Baudrate::B57600 => 57600,
+            EL6021Baudrate::B115200 => 115200,
+        }
+    }
+}
```
1cedsoda: Code is okay. Pleas move the enum to the top and the impls below. Make sure there is an empty line in between.

ethercat-hal/src/devices/el6021.rs
1cedsoda: then use `struct`

ethercat-hal/src/devices/el6021.rs
kraemr: 428e834f00fc5045bd7ea72dc4c69e3cb91b2c6d

ethercat-hal/src/devices/el6021.rs
kraemr: 428e834f00fc5045bd7ea72dc4c69e3cb91b2c6d

ethercat-hal/src/devices/el6021.rs
kraemr: 428e834f00fc5045bd7ea72dc4c69e3cb91b2c6d

ethercat-hal/src/devices/el6021.rs
kraemr: 428e834f00fc5045bd7ea72dc4c69e3cb91b2c6d

common/src/modbus/modbus.rs
kraemr: 9f0771897fcd0915c64401f84b0afb90d036cce2

common/src/modbus/modbus.rs
kraemr: b2a2951cf96240d93859872248df69b48df7f1de

ethercat-hal/src/io/serial_interface.rs
kraemr: hsa_message doesnt pop from a queue, read_message also doesnt. The Inverter actor has queues where data gets pushed and popped

control-core/src/actors/mitsubishi_inverter_rs485.rs
kraemr: 4d7f4da58fb9ea8b9e5d6ac27939f03678549f2e

control-core/src/actors/mitsubishi_inverter_rs485.rs
kraemr: 4f99285dd125a45c898852a16b82dd2696fbcada

ethercat-hal/src/devices/el6021.rs
kraemr: 62ad9870bf137c6d6ec2e98d85bc96fea984da79

ethercat-hal/src/devices/mod.rs
kraemr: d5fdf93fbe56f8781017c8bccf9be815d0758113

ethercat-hal/src/io/serial_interface.rs
kraemr: c1cc40389db533edb17f093c2d324732036bf824

control-core/src/actors/mitsubishi_inverter_rs485.rs
kraemr: 47be30be8d5ac6637c6935de0fd2c4c735d1089a

control-core/src/actors/mitsubishi_inverter_rs485.rs
kraemr: aac23ef2bc1675e2e80907e3a8af3a5fe313bf00

control-core/src/actors/mitsubishi_inverter_rs485.rs
kraemr: 9d5d1de2f1f6ba3e1d76ac98ff9ae10f9851af0c

ethercat-hal/src/io/serial_interface.rs
kraemr: sadly pin::Box bla bla doesnt impl Debug so it doesnt work

control-core/src/actors/mitsubishi_inverter_rs485.rs
kraemr: 17f4f6d8585ca09e166e66d3a2ed5adea483cb74

ethercat-hal/src/devices/el6021.rs
kraemr: explicit_baudrate overrides the baudrate to a custom value if set.

ethercat-hal/src/devices/el6021.rs
kraemr: b992b9c37ef02a0cb6378ca367910c1e086e83ec

ethercat-hal/src/devices/el6021.rs
kraemr: nope

kraemr: removed extended_data_frame as it isnt helpful for el6021

ethercat-hal/src/devices/el6021.rs
kraemr: Since we cant set an explicit rate other than the already available coe values this was removed

kraemr: 94c5903551983e014672766b1a2a4446e496d114

PR #165, opened 2025-05-14
Fix Memory Leak

### LINKED ISSUE #130: Event cacheing uses too much memory
When running the server the memory usage is steadily increasing.
I suppose the serverside event cacheing is responsible for this.

1cedsoda:
fix #130

### DIFF
--- a/control-core/src/socketio/namespace.rs
+++ b/control-core/src/socketio/namespace.rs
@@ -48,7 +48,7 @@ pub trait NamespaceInterface {
     fn cache(
         &mut self,
         event: &GenericEvent,
-        buffer_fn: Box<dyn Fn(&mut Vec<GenericEvent>, &GenericEvent) -> ()>,
+        buffer_fn: &Box<dyn Fn(&mut Vec<GenericEvent>, &GenericEvent) -> ()>,
     );
 
     /// Emits an event to all sockets in the namespace and caches it.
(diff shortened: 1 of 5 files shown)


1cedsoda:
# Memory Usage Optimization: Success!
Excellent improvement! Removing Box::leak() from your PDO derive macros has completely solved the memory growth problem.

### Before
Heap at program end: ~84MB
EtherCAT PDO memory: 80.6MB (95.96% of heap)
Memory continuously growing with each cycle
After (Box::leak removed):

### After
Heap at program end: ~3.4MB (96% reduction!)
Largest allocations:
Panic handling: 2.58MB (75.87% of heap)
Async executor state: ~22KB (0.64% of heap)
What Was Happening
The macro was creating new boxed trait object collections on every get_objects()/get_objects_mut() call and then permanently leaking them with Box::leak(). Since these methods were called in every control loop cycle, your memory grew continuously.

```rust
// The problematic code was:
fn get_objects(&self) -> &[Option<&dyn crate::pdo::RxPdoObject>] {
    let objs = vec![
        // ...field mappings...
    ];
    Box::leak(objs.into_boxed_slice())  // 💥 Memory leak!
}
```

Why It Fixed The Problem
By removing Box::leak(), your code now properly drops these temporary allocations after each call instead of permanently storing them on the heap.

PR #190, opened 2025-05-20
fix for electron build hanging

kraemr:


kraemr:
This fixes it since sandbox false enables us to download the packages that electron needs to work

Issue #320, opened 2025-06-10
Export all Graphs at the same Time to Excel


1cedsoda:
3h

Anatomik5:
fix #324

PR #324, opened 2025-06-12
Group Excel Button

Anatomik5:
issue #320


1cedsoda (changes requested):
Can you move all BigGraph code out of the helpers folder, its not meant for this. Please move it to components/graph

PR #376, opened 2025-06-23
feature-docker-mock

Zabergeier:
Added Dockersupport for release, debug and mock konfigurations, implemented in a docker compose file.


1cedsoda:
Thank you for the changes, we discussed this but it provides no value to us in our development cycle. Thank you for your effort :)

PR #439, opened 2025-06-30
added an integral windup prevention (clamp) for integral error

### LINKED ISSUE #436: Extruder doesnt properly stay in Pressure mode
The core issue probably lies in some sort of state management issue:
- When a user exits the Extruder Page, then at some point the state of the Pressure Pid is lost
- After some time apparently the Pressure Mode stays at one speed

kraemr:
fix #436

### DIFF
--- a/control-core/src/controllers/pid.rs
+++ b/control-core/src/controllers/pid.rs
@@ -18,11 +18,16 @@ pub struct PidController {
     /// Derivative error
     ed: f64,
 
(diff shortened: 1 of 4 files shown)


kraemr:
The Extruder being stuck at one speed then recovering most likely is caused by integral windup in the PIDController. By clamping integral error to a reasonable max and min value this can be avoided

PR #460, opened 2025-07-01
refactor: Run EtherCAT initialization in a separate thread

1cedsoda:
This change refactors the EtherCAT initialization process to be non-blocking by moving it into its own dedicated thread.

- The `init_ethercat` function now spawns a named thread ("init_ethercat") to handle interface discovery and the setup loop.
- This prevents the main application from blocking during startup while searching for a valid EtherCAT interface.
- The initialization order in `main.rs` has been slightly adjusted.
- A minor optimization was made to the laser machine actor to avoid redundant `Instant::now()` calls.

### DIFF
--- a/server/src/ethercat/init.rs
+++ b/server/src/ethercat/init.rs
@@ -1,7 +1,7 @@
 use super::setup::setup_loop;
 use crate::{
     app_state::AppState,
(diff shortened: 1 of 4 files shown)


1cedsoda:
I yanked this from master again

PR #480, opened 2025-07-07
471 add winder auto stop or pull after n meters of filament

### LINKED ISSUE #471: Add Winder Auto Stop or Pull after N meters of filament
- Add a checkbox enabling or disabling it
- Two buttons to choose between Standby or Pull after N Meters
- Edit the Meters of Filament
- Backend needs to keep track of the meters of filament pulled

calculation for meters of filament?: 
7.9 cm diameter for the rolls
perimeter * RPM

kraemr:
fix #471


kraemr:
Not sure how to name the variables here

PR #504, opened 2025-07-17
Winder 1 docs v2

### LINKED ISSUE #64: Handbook for Winder V2

kraemr:
fix #64


kraemr:
The Markdown rendering is completely broken. When you use the ManualPage of the Winder it doesnt render any images or <br> and similiar things. The Markdown file itself is perfectly fine and renders as expected when rendered in VScode, Browser or Github

kraemr:
So i just looked a little deeper into the markdown file. It uses some non-standard stuff like `<br>` and `` which is technically html but works fine in other renderers. Also the images that are already supplied with a markdown link  dont load, perhaps due to it being a relative path? Check if an absoulute Path fixes it.

TheBest6337:
grafik

So there's one annoying issue left, where the table formatting is completely weird. On some, the sizing is off for no apparent reason, and for others, the columns are misaligned. I'm currently not sure how to fix it as I don't know the cause yet, but it's fixable if the tables would be combined.

grafik

PR #505, opened 2025-07-17
Build front-end in CI

Oshgnacknak:
Currently, electron-forge seems to mute a lot of the error messages.
This is why first compiling with tsc might help


kraemr:
@Oshgnacknak Fix the compile errors please, then you can merge. Probably because of rebase

Issue #62, opened 2025-04-11
Persist Config Values


1cedsoda:
We need a way to persists config and other values in the front-end.

Requirements:

- [ ] Presets are versioned and use migrations
  - [ ] Presets are persisted for each version (to not break preset in case of a version rollback)
- [x] Presets are persisted in the locally
- [ ] Presets can be imported/exported to USB drive
  - [ ] A `undefined` config value while importing gets overridden with the default value
- [ ] Each Machine is has "presets"
  - [x] A preset that matches the current state is marked as "active"
  - [x] Many inactive Preset
  - [x] Presets can be saved with a name, overridden, deleted
  - [x] `null` is an explicitly empty value
  - [x] Before applying a preset, User must see which values are modified
    - [x] including values that are reset to their default
  - [x] The user must be aware what the defaults are
    - [x] A "Default" Preset is rendered as a read only option in the list
  - [x] The "Latest" preset is a read only preset that gets overridden after each user input
  - [ ] Laser 1
  - [x] Mock 1
  - [x] Extruder 2
  - [x] Winder 2

Oshgnacknak:
The remaining tasks have been moved to their own issues.
Closed with #514

PR #514, opened 2025-07-21
Persist machine configuration in presets

Oshgnacknak:
#62

As discussed, the following features are not implemented yet but will become their own issue:

- Checking the schema version of a preset and possibly running migrations of needed.
- Being able to export presets into an USB stick.
- Implementing presets for the Laser.


.github/workflows/electron.yml
kraemr, line 1: Open a different Issue for this

electron/package.json
kraemr, line 1: Open a different Issue for this

electron/src/machines/mock/mock1/Mock1Graph.tsx
kraemr, line 1: Why did you rename sineWave to sineWaveSum here ?

tmux-dev.sh
kraemr, line 1: This does not belong in here, remove it

electron/src/components/preset/PresetPreviewTable.tsx
```diff
@@ -0,0 +1,58 @@
+import React, { Fragment } from "react";
+import { PresetData, PresetSchema } from "@/lib/preset/preset";
+import { renderUnitSymbol, renderUnitSyntax, Unit } from "@/control/units";
+
+export const previewSeparator = undefined;
```
kraemr, line 5: Why do you have a const value that is always undefined? Just use undefined

electron/src/components/preset/PresetPreviewTable.tsx
```diff
@@ -0,0 +1,58 @@
...
+};
+
+export function PresetPreviewTable<T extends PresetSchema>({
+  data,
+  entries,
+}: PresetPreviewTableProps<T>) {
+  return (
+    <div className="grid grid-cols-[60%_10%_30%]">
+      {entries.map((entry, i) => {
+        if (entry === previewSeparator) {
```
kraemr, line 32: just do === undefined here

electron/src/lib/preset/preset.ts
```diff
@@ -0,0 +1,20 @@
+import { machineIdentificaiton } from "@/machines/types";
+import { z } from "zod";
+
+export type PresetSchema = z.ZodTypeAny;
+
+export const presetSchema = <S extends PresetSchema>(dataSchema: S) =>
+  z.object({
+    id: z.number().int().nonnegative().optional(),
```
kraemr, line 8: Why can a preset id be optional (undefined)?

electron/src/lib/preset/presetStore.ts
kraemr, line 1: 'PersistedState' is defined but never used. Will this be used later on?

electron/src/lib/preset/usePresets.ts
kraemr, line 1: Remove the TODO Comments and write what you want to do in a ticket instead

electron/src/components/preset/PresetPreviewTable.tsx
Oshgnacknak: Because we might have to change this later, then we don't need to do this in multiple places - or at least, the compiler tells us where.

Currently undefined has all the information we need. It is possible that this might change in the future. Any unique value (string, symbols, `null`, `undefined`) works. `undefined` is just the most light weight.

Also on the other side, if you read `undefined` in the list. Would you be able to guess what effect this has? Using a name here makes it obvious what the desired behavior should be.

electron/src/components/preset/PresetPreviewTable.tsx
Oshgnacknak: See my comment on the above code

electron/src/lib/preset/preset.ts
Oshgnacknak: Because Presets that don't come directly from the Preset Store (imported by the user, or just an dummy value to give to pass around in the ui) might not have an id. The id only makes sense if a preset is directly handled the store.

If we force an id everywhere, sometimes it will have an incorrect dummy value, which I consider more confusing that declaring the field as optional.

Its true that one could also create a type Representing only the parts of a preset the user is directly interested in (name, date, data), that is used instead. Any opinion on this?

electron/src/lib/preset/presetStore.ts
Oshgnacknak: Here its up to you, do you want me to immediately declare the type of a schema, or only once we need it.

For now, I just made them used, which enhanced type safety.

electron/src/lib/preset/preset.ts
kraemr: Alright then that makes sense.We will keep it that way for now

PR #515, opened 2025-07-22
Fix typo in machineIdentification

Oshgnacknak:


kraemr:
typo is already fixed on main

PR #517, opened 2025-07-22
482 communicate with winder

### LINKED ISSUE #482: Communicate with Winder
The Buffer needs to be able to communicate with the Winder.
On the Winder frontend a user should be able to start buffering with a button for example.

Need more info!

Snacj:
fix #482

### DIFF
--- a/electron/src/lib/preset/preset.ts
+++ b/electron/src/lib/preset/preset.ts
@@ -1,4 +1,4 @@
-import { machineIdentificaiton } from "@/machines/types";
+import { machineIdentification } from "@/machines/types";
 import { z } from "zod";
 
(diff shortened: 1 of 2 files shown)


cargo_run_linux.sh
kraemr, line 1: Remove this

electron/src/machines/winder/winder2/winder2Namespace.ts
```diff
@@ -22,6 +22,7 @@ import {
   TimeSeries,
   TimeSeriesValue,
 } from "@/lib/timeseries";
+import { connected } from "process";
```
kraemr, line 25: Is defined but never used

PR #551, opened 2025-08-04
550 allow wound distance reset during winding

TheBest6337:
Adds the functionality of resetting the wound distance even when the winding process is running. When running a dialog appears as a warning.


kraemr (approved):
LGTM

PR #592, opened 2025-08-14
Fix 540: long preset names mess with ui

### LINKED ISSUE #540: Long preset names mess with UI
One could limit the name's length. Also possible to fix this in the ui, either with linebreaks or with ellipses (...).

Need a comment how this should be fixed

Oshgnacknak:
Long names are now shorted with an ellipsis.
Other options are in a hamburger menu to save space.

fixes #540


1cedsoda (approved):
yallah

Issue #591, opened 2025-08-14
Get i915 driver to load properly


kraemr:
Currently on the nixos system one Frame takes 38ms above our target of 33ms.
Problem seems to be, that electron uses llvmpipe (software-rednering) instead of the intel gfx

Solutions:
- Get i915 driver to load under PREEMPT_RT kernel

kraemr:
After running glxinfo | grep "OpenGL" on the new preempt_rt kernel it shows llvmpipe, which in essence means everything is software rendered causing these horrendous frame times

PR #609, opened 2025-08-20
597 use latest rt nixos kernel

kraemr:


kraemr:
already using the latest since #591 was closed

Issue #612, opened 2025-08-20
Implement connection between winder and laser


Snacj:

TheBest6337:
@Snacj whats the status on this?

Snacj:
@TheBest6337 It's been on hold since the machine connection was broken.

TheBest6337:
progress was made and we will probably be finished with this feature today.

PR #625, opened 2025-08-26
Introduce machine connection

### LINKED ISSUE #559: Build a MachineWrapper with Caching
When Machines disconnect (Laser included), then all data is lost. To solve this issue we need to decouple the namespaces and machines from another

Oshgnacknak:
This is to hold on to the namespace of
a machine even if the devices of that
machine have disconnected and to allow
reattaching to the namespace once the
devices reconnects.

Additionally to the new struct,
a vast refactor took place cleaning up
connections to machines, cross connections
between machines the way we downcast
machines downcasts.

Fixed #559

### DIFF
--- a/server/src/machines/aquapath1/mod.rs
+++ b/server/src/machines/aquapath1/mod.rs
@@ -1,11 +1,9 @@
 use control_core::{
-    machines::{
-        Machine,
-        identification::{MachineIdentification, MachineIdentificationUnique},
-    },
+    machines::identification::{MachineIdentification, MachineIdentificationUnique},
     socketio::namespace::NamespaceCacheingLogic,
 };
 
+use control_core_derive::Machine;
 use serde::{Deserialize, Serialize};
 use std::time::Instant;
 use uom::si::{
@@ -80,7 +78,7 @@ impl Default for Flow {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Machine)]
 pub struct AquaPathV1 {
     machine_identification_unique: MachineIdentificationUnique,
     namespace: AquaPathV1Namespace,
@@ -89,17 +87,12 @@ pub struct AquaPathV1 {
     front_controller: Controller,
     back_controller: Controller,
 }
+
 impl AquaPathV1 {
     pub const MACHINE_IDENTIFICATION: MachineIdentification = MachineIdentification {
         vendor: VENDOR_QITECH,
         machine: MACHINE_AQUAPATH_V1,
     };
-}
-
-impl Machine for AquaPathV1 {
-    fn get_machine_identification_unique(&self) -> MachineIdentificationUnique {
-        self.machine_identification_unique.clone()
-    }
 }
 
 impl std::fmt::Display for AquaPathV1 {
--- a/server/src/machines/extruder1/mock/mod.rs
+++ b/server/src/machines/extruder1/mock/mod.rs
@@ -1,9 +1,9 @@
 #[cfg(feature = "mock-machine")]
-use control_core::machines::Machine;
-#[cfg(feature = "mock-machine")]
 use control_core::machines::identification::MachineIdentification;
 #[cfg(feature = "mock-machine")]
 use control_core::machines::identification::MachineIdentificationUnique;
+#[cfg(feature = "mock-machine")]
+use control_core_derive::Machine;
 #[cfg(feature = "mock-machine")]
 use std::time::Instant;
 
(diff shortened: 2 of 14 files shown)


Oshgnacknak:
This also changes the way #612 is implemented. @Snacj

Oshgnacknak:
Held as draft because the laser seems to only connect after being plugged out and in again.

Oshgnacknak:
Closing and reopening frontend made the whole thing buggy. Backend kept running but forntend could not get machine states and I had to restart the whole machine.
This could be because if the refactor

Oshgnacknak:
last issue seems to be gone after rebase

Oshgnacknak:
Last force push after fixing laser not connected.

control-core/src/machines/connection.rs
```diff
@@ -0,0 +1,205 @@
...
+
+#[derive(Debug)]
+pub struct MachineCrossConnection<
+    F: CrossConnectableMachine<F, T>,
+    T: CrossConnectableMachine<T, F>,
+> {
+    machine_manager: Weak<RwLock<MachineManager>>,
+    machine_identification_unique: MachineIdentificationUnique,
+    connected_machine: Weak<Mutex<MachineSlot<T>>>,
+    _make_compiler_happy: std::marker::PhantomData<F>,
```
kraemr: whats __make_compiler_happy for? i mean its in the name but why does the compiler complain if its not there? Add a comment why its there

kraemr:
So the Code looks good mostly, but this will not go into the next release, it will go into the one after. Also before this is merged into main id like to test this in prod

server/src/socketio/main_namespace/machines_event.rs
```diff
@@ -22,15 +22,14 @@ impl MachinesEventBuilder {
...
         for machine in machines_guard.iter() {
+            let slot = machine.1.lock_blocking();
+            let connection = &slot.machine_connection;
             machine_objs.push(MachineObj {
                 machine_identification_unique: machine.0.clone(),
-                error: match machine.1 {
-                    Ok(_) => None,
-                    Err(e) => Some(e.to_string()),
-                },
+                error: connection.to_error().map(|e| e.to_string()),
```
kraemr: Remove the TODO or if you plan to impl it ... do it

kraemr (changes requested):

control-core/src/machines/connection.rs
Oshgnacknak: We want to hint the TypeChecker that this is a connection from F to T, hence the letters. Unless somewhere, F is used, the compiler will complain about an unused type argument. However, PhantomData tells the compiler, that we want it to assume that F is being used, even if only for type information.

Without F, we would not be able to go back to ourselfs using `self.get_cross_connection().get_cross_connection()`. We would lose type information that way. This then would make it impossible to properly implement `reverse_connect`.

Oshgnacknak:
> So the Code looks good mostly, but this will not go into the next release, it will go into the one after. Also before this is merged into main id like to test this in prod

Tested it on our production line. Works like a charm.

control-core/src/machines/connection.rs
kraemr: Makes sense i guess? Just add a comment alluding to that

server/src/socketio/main_namespace/machines_event.rs
Oshgnacknak: Cleaned up quite this part quite a bit

kraemr:
Looks good, resolve conflicts then ill approve

server/src/machines/aquapath1/mod.rs
```diff
@@ -77,8 +80,9 @@ impl Default for Flow {
     }
 }
```
kraemr, line 80: Add the machine derive back again

server/src/machines/extruder1/mock/mod.rs
```diff
@@ -29,8 +31,9 @@ pub mod mock_emit;
 pub mod new;
 
 #[cfg(feature = "mock-machine")]
-#[derive(Debug, Machine)]
+#[derive(Debug)]
 pub struct ExtruderV2 {
+    machine_identification_unique: MachineIdentificationUnique,
     namespace: ExtruderV2Namespace,
     last_measurement_emit: Instant,
     last_status_hash: Option<u64>,
```
kraemr, line 39: Also here, why did you remove Machine derive

kraemr (changes requested):
i think in winder Machine derive is missing add it back, then it is fine. also dw i wont merge anything for now :)

server/src/machines/extruder1/mock/mod.rs
kraemr: Oh wait its in mock thats fine then nvm

Oshgnacknak:
@kraemr trait Machine is being derived again. I checked, when missing the field, the error message is very understandable.

kraemr (approved):

PR #639, opened 2025-08-28
Add reverse prop to `SelectionGroupBoolean`

1cedsoda:
This change introduces a `reverse` property to the `SelectionGroupBoolean` component to allow for more flexible UI representation.

- When `reverse` is set to true, the `optionTrue` and `optionFalse` props are swapped.
- This allows the visual "on" state to correspond to a `false` value and vice-versa.
- The prop is implemented on the Inverter Regulation control on the Extruder2 page.


kraemr (approved):
LGTM

PR #657, opened 2025-09-01
added global timeframe store

### LINKED ISSUE #634: store Graphs timeframe globally
currently the timeframe, 30min,1h etc is not saved in the global state. Which means that the graph page forgets all its settings

kraemr:
fix #634


1cedsoda (approved):
works i guess

PR #659, opened 2025-09-02
575 automatically send stateevent when changed

### LINKED ISSUE #575: Automatically send StateEvent when changed
Currently it can happen that the state mismatches the actual state, especially for the extruder, as extruder errors are considered state.

kraemr:
fix #575


1cedsoda (approved):
nice

PR #663, opened 2025-09-03
521 retrieve detailed faults for the inverter

### LINKED ISSUE #521: Retrieve detailed faults for the inverter
Currently we only check the status of the inverter. We should also check the specific inverter fault.
These can seemingly be read from the fault history.

kraemr:
fix #521


1cedsoda (changes requested):
- [ ] Test on device
- [ ] Resolve conflicts

Oshgnacknak:
@kraemr Whats state on this?

kraemr:
Wasnt really a priority, which is why its rotting in here as a draft. Probably need to Redo it anyways since there are two Extruder versions / Inverters used and its really old. Closing

PR #677, opened 2025-09-07
676 reduce heap allocationsdeallocations in machineact

### LINKED ISSUE #676: Reduce Heap Allocations/Deallocations in MachineAct
Currently in every impl of MachineAct we use a Box::new(asyc move ... ), this in turn allocates memory on the heap and deallocates it later on, again. This most likely contributes to inconsistent timing in our rt sensitive loops. 

idea: 
```rust
impl MachineAct for ExtruderV2 {
    fn act(&mut self, now: Instant) -> Pin<&mut (dyn Future<Output = ()> + Send + '_)> {
        // refresh the slot so it's a "completed" Ready future again
        self.future_slot = future::ready(());
        // return pinned &mut reference as a trait object
        Pin::new(&mut self.future_slot) as Pin<&mut (dyn Future<Output = ()> + Send + '_)>
    }
}
```
instead of creating a new future every time the loop is called we just save a future inside of each machine. This would in theory only allocate a future on startup and shutdown, while just giving a pinned reference to the local future on the struct. 

This solution would also not completely destroy our dyn Machine trait logic, which would be neat

kraemr:
fix #676


kraemr:
Need to test this with the machines, atleast it compiles.

PR #684, opened 2025-09-08
683 ethercat loop broken

### LINKED ISSUE #683: Ethercat Loop broken
Seems like Ethercat Output is broken due to timing mismatch, perhaps just sleep 1ms after the inputs are copied?

kraemr:
fix #683


kraemr:
We now use a static 500 microseconds delay instead the overly complex LoopThrottle

PR #698, opened 2025-09-10
696 pin ethernet interrupt handler to the same core as ethercrab task

### LINKED ISSUE #696: Pin Ethernet Interrupt Handler to the same Core as Ethercrab Task
This heavily improved the 99.99th percentile measurements.
```bash
cat /proc/interrupts | grep interface_name
# RUN NIC IRQ on Same RT Thread
echo 4 > /proc/irq/<irq_number>/smp_affinity_list

```
After Interface Discovery make sure that the interface is pinned to the correct core.

kraemr:
fix #696

### DIFF
--- a/control-core/src/irq_handling.rs
+++ b/control-core/src/irq_handling.rs
@@ -49,19 +49,21 @@ fn get_interface_irq(proc_content: &str, interface_name: &str) -> Option<u32> {
     None
 }
 
-// Since kernel 3.0 it’s possible to use the /proc/irq/<IRQ-NUMBER>/smp_affinity_list
-// With comma seperated values
+/// Since kernel 3.0 it’s possible to use the /proc/irq/<IRQ-NUMBER>/smp_affinity_list
+/// With comma seperated values
+/// This function takes the irq identifier and writes the cpu string
+/// into /proc/irq/irq_number/smp_affinity_list
 #[cfg(unix)]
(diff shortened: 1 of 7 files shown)


control-core/src/realtime.rs
```diff
@@ -1,6 +1,8 @@
 use tracing::debug;
 use tracing::error;
 
+use libc;
```
1cedsoda: remove this

server/src/ethercat/setup.rs
```diff
@@ -44,6 +45,11 @@ pub async fn setup_loop(
         .name("EthercatTxRxThread".to_owned())
         .spawn(move || {
             send_panic(thread_panic_tx_clone);
+            if let Err(e) = set_irq_handler_affinity(&interface, 3) {
```
1cedsoda: missing feature flag

control-core/src/irq_handling.rs
```diff
@@ -0,0 +1,103 @@
+use std::ffi::CString;
```
1cedsoda, line 4: missing libc import?

1cedsoda (changes requested):

control-core/src/realtime.rs
kraemr: ok boss

server/src/ethercat/setup.rs
kraemr: i set it boss

PR #704, opened 2025-09-10
Remove craneLib from server nix package

Oshgnacknak:


kraemr:
@Oshgnacknak  can i get a reason as to why you want to remove it?

nixos/os/configuration.nix
```diff
@@ -315,7 +315,7 @@
   # List services that you want to enable:
 
   # Enable the OpenSSH daemon.
-  # services.openssh.enable = true;
+  services.openssh.enable = true;
```
kraemr: Rmove this should be false

nixos/packages/server.nix
```diff
@@ -1,66 +1,38 @@
...
 
-  postBuild = ''
-    if [ -f $TMPDIR/swap/swapfile ]; then
-      swapoff $TMPDIR/swap/swapfile
-      rm $TMPDIR/swap/swapfile
-    fi
-  '';
+  JEMALLOC_SYS_WITHOUT_VENDOR = "1";
+  CFLAGS = "-Wno-error=int-conversion";
+  CXXFLAGS = "-Wno-error=int-conversion";
```
kraemr: leave prbuild and postbuild as is

Oshgnacknak:
Removing craneLib already speeds up compilation time a little bit (imo).

Also, we can pin dependencies using `cargoLock.outputHashes' later on.
https://artemis.sh/2023/07/08/nix-rust-project-with-git-dependencies.html

kraemr:
@Oshgnacknak  Resolve conflicts and test again if compilation is 1. successful, 2. faster perhaps?

Oshgnacknak:
@kraemr got it building again. However, seems slower them before.

kraemr:
> @kraemr got it building again. However, seems slower them before.

If its slower then there is no point. Will close then.

PR #705, opened 2025-09-10
Disable hardening in nix develop shell

Oshgnacknak:
Hardening causes jmalloc to fail at compilation


kraemr:
@Oshgnacknak Ok Cool but what impact does disabling this have?           hardeningDisable = [ "fortify" ];

Oshgnacknak:
It add compiler flags, that conflict with how jmalloc needs to be build.
See https://github.com/NixOS/nixpkgs/issues/18995

PR #711, opened 2025-09-11
added collect-garbage on each delete

### LINKED ISSUE #694: Over Time NixOs Generations takes up all Diskspace even when Deleted
Apparently Generations leave behind a lot of garbage that needs to be manually cleaned up. Every time a generation is deleted either run the command to delete all garbage or do it specifically for that version (not sure if possible). 

https://nix.dev/manual/nix/2.18/command-ref/nix-collect-garbage

kraemr:
fix #694


kraemr:
Need to test if this actually only removes dangling store paths.

PR #712, opened 2025-09-11
Read git information during build instead of addign new commit via installInfo.nix

Oshgnacknak:
Problem is that the new commit corrupts hashes and therefore makes caching impossible. Might also be responsible for the builds taking forever.

~~Currently this is done quite hacky. In fact, it is not possible to get access to the `.git` folder to access this information. Git cannot add its own folder and therefore nixos will never copy these files over during `nix-rebuild`.~~

While we still need the install script, now no new commit needs to be added to the history. Also the no janky copy in this new approach.

### DIFF
--- a/cargo_run_linux.sh
+++ b/cargo_run_linux.sh
@@ -24,7 +24,7 @@ if [ "$1" != "mock-machine" ]; then
     sudo setcap 'cap_net_raw,cap_sys_nice=eip' ./target/debug/server
 
     echo "Setting permissions for serial ports"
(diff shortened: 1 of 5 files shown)


kraemr:
@Oshgnacknak currently dont have a nix machine at home. Will test on Monday, looks like it should work though.

kraemr:
~~Actually you also need to rewrite some frontend logic, because the Update site doesnt display the git information correctly. on that branch.~~

kraemr (dismissed):

kraemr (approved):

PR #725, opened 2025-09-15
425 fix terminal component issues v2

### LINKED ISSUE #425: Fix `Terminal` Component Issues
- When using a terminal component multiple times (like multiple updates) the lines are replicated n times
- Use list virtualization to improve performance for long logs

kraemr:
fix #425

### DIFF
--- a/electron/src/helpers/update_helpers.ts
+++ b/electron/src/helpers/update_helpers.ts
@@ -1,16 +1,28 @@
 import type { UpdateInfo } from "@/stores/updateStore";
 import { useUpdateStore } from "@/stores/updateStore";
 
+let currentLogListener: ((line: string) => void) | null = null;
+
 export async function updateExecute(
   source: UpdateInfo,
   onLog: (log: string) => void,
 ): Promise<{ success: boolean; error?: string }> {
   return new Promise((resolve) => {
-    window.update.onLog(onLog);
+    // Remove previous listener if exists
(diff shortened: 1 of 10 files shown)


electron/src/helpers/update_helpers.ts
```diff
@@ -1,16 +1,28 @@
...
+    }
+
+    currentLogListener = onLog;
+    window.update.onLog(currentLogListener);
+
     window.update.execute(source);
+
     window.update.onEnd((params) => {
-      window.update.onLog(() => {});
+      window.update.onLog(() => {}); // remove listener
```
Oshgnacknak, line 23: Optimally, there would be a method called `removeOnLogListener` or something like that. But thats just the icing on the cake.

nixos/modules/qitech.nix
```diff
@@ -107,7 +107,7 @@ serviceConfig = {
       
       environment = {
         RUST_BACKTRACE = "full";
-        RUST_LOG = "info";
+        RUST_LOG = "debug";
```
Oshgnacknak: I assume we don't want this on master

Oshgnacknak (changes requested):

nixos/modules/qitech.nix
kraemr: We had this on master a long time as a default but yeah id say lets stick with info

PR #749, opened 2025-09-24
738 add mock machines

### LINKED ISSUE #738: Add ExtruderV2 Mock-Machine
Leverage Traits and the existing mock-machine feature create mock-machines with as little code as possible. In essence we can just use the existing Extruder-site for example, but use the mock-extruder backend, which is basically the same, but doesnt require any hardware to be attached

kraemr:
fix #738

### DIFF
--- a/server/src/machines/extruder1/mock/mock_emit.rs
+++ b/server/src/machines/extruder1/mock/mock_emit.rs
@@ -1,0 +1,156 @@
+// imitates the api behaviours of an ExtruderV2
+
+use control_core::{
+    helpers::hasher_serializer::hash_with_serde_model,
+    socketio::{event::BuildEvent, namespace::NamespaceCacheingLogic},
+};
+
+use crate::machines::extruder1::{
+    ExtruderV2Mode, HeatingType,
+    api::{ExtruderV2Events, LiveValuesEvent, PidSettings, StateEvent},
+    mock::ExtruderV2,
+};
+
+//#[cfg(feature = "mock-machine")]
+impl ExtruderV2 {
+    pub fn build_state_event(&mut self) -> StateEvent {
+        // bad performance wise, but doesnt matter its only a mock machine
+        StateEvent {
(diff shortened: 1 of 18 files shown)


server/src/machines/extruder1/mock/new.rs
```diff
@@ -0,0 +1,133 @@
...
+        InverterStatusState, ModeState, MotorStatusValues, PidSettings, PidSettingsStates,
+        PressureState, RegulationState, RotationState, ScrewState,
+    },
+    mock::ExtruderV2,
+};
+
+use anyhow::Error;
+use control_core::machines::new::{MachineNewHardware, MachineNewTrait};
+
+//#[cfg(feature = "mock-machine")]
```
Oshgnacknak: sure this should be commented out?

server/src/machines/extruder1/mock/mock_emit.rs
```diff
@@ -0,0 +1,156 @@
...
+            HeatingType::Back => self.heating_states.back.target_temperature = target_temperature,
+            HeatingType::Middle => {
+                self.heating_states.middle.target_temperature = target_temperature
+            }
+        }
+        self.emit_state();
+    }
+
+    pub fn configure_pressure_pid(&mut self, settings: PidSettings) {
+        // self.screw_speed_controller
```
Oshgnacknak: Should this still exist or is it old code?

Oshgnacknak (changes requested):
Mostly looks nice, however, there still is some commented out code, which really reduces readability.

server/src/machines/extruder1/mock/new.rs
kraemr: Nope i think not. Good that you caught that. I think it still works, but the flag makes it clearer

server/src/machines/extruder1/mock/mock_emit.rs
kraemr: old code, will remove

Oshgnacknak (approved):

PR #754, opened 2025-09-25
748 add feature flag or env var for setting loop timeout

### LINKED ISSUE #748: Add feature flag or env var for setting loop timeout
Why? Because on dev machines we cannot achive the same realtime performance as with an rt kernle, so we need to be able to set it manually either through an env var or feature flag. Most general purpose machines get ~1ms tx rx time

kraemr:
fix #748


kraemr:
@TheBest6337 Please try compiling on your system.

PR #776, opened 2025-10-03
Add padding parameter explanation to manual for filament winding

TheBest6337:


TheBest6337:
Milan wants me to add an illustration to it, so im marking it as a draft.

PR #810, opened 2025-10-15
Use Linux source code from realtime kernel 6.12

Oshgnacknak:
We don't plan a kernel update after some consideration. Nevertheless this pull request serves as a starting point for a potential future update.


Oshgnacknak:
Subverted by #829

PR #847, opened 2025-10-22
started refactoring initialization to use futures instead

### LINKED ISSUE #846: Refactor Initialization to be based around futures
Instead of starting MULTIPLE threads that run the initialization functions, where the threads themselves ALSO use async BUT BLOCK ON IT!?!?!!? just make these init functions futures that return the found "devices" or "interfaces".

Idea: 
Instead of: 
```
pub fn init_serial(app_state: Arc<AppState>) -> Result<(), anyhow::Error> {
    let app_state_clone = app_state.clone();

    thread::Builder::new()
        .name("SerialDetectionThread".to_owned())
        .spawn(move || {
            let app_state_clone = app_state_clone;
            smol::block_on(async move {
                let rt = smol::LocalExecutor::new();
                rt.run(async {
                    loop {
                        let result = {
                            let mut serial_setup_guard = app_state_clone.serial_setup.write().await;

                            let mut port_result =
                                serial_setup_guard.serial_detection.check_ports().await;

                            let mut removed_signals = serial_setup_guard
                                .serial_detection
                                .check_remove_signals()
                                .await;

                            port_result.removed.append(&mut removed_signals);

                            port_result
                        };

                        if !result.added.is_empty() || !result.removed.is_empty() {
                            // sync serial device discovery to machine manager
                            {
                                let mut machine_guard = app_state_clone.machines.write().await;
                                // turn added devices into machines
                                for (device_identifiaction, device) in result.added {
                                    machine_guard.add_serial_device(
                                        &device_identifiaction,
                                        device.clone(),
                                        &MACHINE_REGISTRY,
                                        app_state_clone.socketio_setup.socket_queue_tx.clone(),
                                        Arc::downgrade(&app_state.machines),
                                    )
                                }
                                for device_identification in result.removed {
                                    machine_guard.remove_serial_device(&device_identification)
                                }
                            }

                            // Notify client via socketio
                            let app_state_event = app_state.clone();
                            smol::block_on(async {
                                let main_namespace = &mut app_state_event
                                    .socketio_setup
                                    .namespaces
                                    .write()
                                    .await
                                    .main_namespace;
                                let event = MachinesEventBuilder().build(app_state_event.clone());
                                main_namespace.emit(MainNamespaceEvents::MachinesEvent(event));
                            });
                        }
                        smol::Timer::after(Duration::from_millis(300)).await;
                    }
                })
                .await;
            });
        })
        .expect("Failed to spawn SerialTxRxThread");
    Ok(())
}
```

We simply do something like this: 
```
use crate::socketio::main_namespace::MainNamespaceEvents;
use crate::socketio::main_namespace::machines_event::MachinesEventBuilder;
use crate::{app_state::AppState, machines::registry::MACHINE_REGISTRY};
use anyhow::Result;
use control_core::serial::SerialDeviceIdentification;
use control_core::serial::serial_detection::SerialDetection;
use control_core::socketio::namespace::NamespaceCacheingLogic;
use serialport::UsbPortInfo;
use smol::Timer;
use std::{sync::Arc, thread, time::Duration};

// Returns a completed future, when atleast ONE serial devices were found

pub fn find_serial_future()
-> impl future::Future<Output = HashMap<SerialDeviceIdentification, UsbPortInfo>> {
    future::lazy(|_| async {
        loop {
            let devices = SerialDetection::detect_devices().await;
            if !devices.is_empty() {
                return devices;
            }
            Timer::after(Duration::from_secs(1)).await;
        }
    })
}

```

In main we can then simply poll this function until it completes OR we can continue, it depends on the machines and what they require. For our Framework we want as little hardcoded as possible, for example our Machine registry etc, we would rip all that stuff out, because this should be handled by the user of the library, also the frontend would be ripped out as well.

kraemr:
fix #846

### DIFF
--- a/server/src/serial/init.rs
+++ b/server/src/serial/init.rs
@@ -94,7 +94,6 @@ pub async fn handle_serial_device_hotplug(
             _ => (),
         };
     } else if laser.is_none() && unique_ident.is_some() {
-        tracing::info!("hello non existing laser");
         let serial_params = SerialDeviceNewParams {
             path: "".to_string(),
         };
@@ -131,7 +130,6 @@ pub async fn handle_serial_device_hotplug(
                     .await
                     .main_namespace;
 
-                tracing::info!("{:?}", app_state.machines);
                 let event = MachinesEventBuilder().build(app_state.clone());
                 main_namespace.emit(MainNamespaceEvents::MachinesEvent(event));
             }


kraemr:
Runtime Cost of the init threads etc, is now VERY negligible, i get 0.0 - 1.3% CPU-Usage with all async tasks running and processes.  Still need to add back EtherCatDevice events, and move the serial detection out into some server module.

kraemr:
@Oshgnacknak If you are in the office please test this on a real machine before merging

Oshgnacknak:
@kraemr Code looks good. 

Tested in now. Ended up timeouting at some point. Not sure if futures caused this, or if the logging is just misleading. Log is attached. The cicle times look really consistent however, so I would be surprised if it was something like a kernel hickup.

[qitech_control_server_journald_2025-10-24T16-07-55.log](https://github.com/user-attachments/files/23128620/qitech_control_server_journald_2025-10-24T16-07-55.log)

kraemr:
hmm interesting, perhaps i messed up with all the locking? The cycle times do look better now, probably because there is less context switching and more cache locality i think. Maybe all the async i/o lands on lands on the same async/io thread regardless? 

In theory the for futures only init and socketio and serialdiscovery run continuously in the main thread. Maybe the smol::block_on() + locking in our server loop causes us issues? Because EtherCat requires deterministic timing

kraemr:
> @kraemr Code looks good.
> 
> Tested in now. Ended up timeouting at some point. Not sure if futures caused this, or if the logging is just misleading. Log is attached. The cicle times look really consistent however, so I would be surprised if it was something like a kernel hickup.
> 
> [qitech_control_server_journald_2025-10-24T16-07-55.log](https://github.com/user-attachments/files/23128620/qitech_control_server_journald_2025-10-24T16-07-55.log)

Im pretty sure it is due to the fact, that we have no timeout anymore in the server loop, basically a busy loop, now in that busyloop for EVERY cycle we instantiated a new smol async task, which most likely caused a Timeout at some point. Now that we instantiate only once the times have also gotten better. If this reappears then we will just add a spin_loop at the end to make sure we hit our timings. Funnily enough almost all the locks are just wasted time, since only the server loop continuously accesses the devices and machines

PR #849, opened 2025-10-24
769 add an alert or warning in the frontend for filament out of spec

### LINKED ISSUE #769: Add an Alert or Warning in the frontend for filament out of spec
When a filament is out of spec either by A LOT or a little bit for a long time throw up a warning or alert and maybe also add a checkbox where when it is ticked the winder and extruder auto stop.

Snacj:
fix #769


Snacj:
@kraemr do NOT include in release until laser - winder connection is fully established and tested. It is needed to include the option to disable the winder automatically if the laser is out of tolerance.

Snacj:
I implemented and updated the Laser Alert if its out of tolerance it the backend. This can be merged. I will make a new Issue to determine machine stops via machine connections

PR #911, opened 2025-11-19
909 implement ecmy error messages

### LINKED ISSUE #909: Implement ECMA error messages
In Ethercat Couplers there is a specific register, that contains an error message in ECMA format (binary). We should implement a parser for this and print the "pretty" messages, so we always know why a coupler for example didn't enter OP or PRE-OP.

kraemr:
fix #909


Snacj:
MAX_PDU_DATA and PDI_LEN differ between *ethercat-eeprom-dump/src/main.rs* and *server/src/ethercat/config.rs* (512/1024).
Is this desired?

PR #915, opened 2025-11-22
Implement Wago Power Supply Machine

Oshgnacknak:
In addition to implementing the front-end and back-end logic, this machine is not initialized in our default way. Their is no need need to implement `MachineNewTrait` as the information provided by `MachineNewParams` only gets used by EtherCAT Machines in the first place. `MachineNewTrait` was dropped from the `Machine` trait.

Instead of using the machine registry - which only stores closures anyway - we just create a Machine without a namespace that only knows how to talk to the hardware it controls and to the shared state via channels. Hence, a new struct `MachineChannel` with a default impl for `Machine` for anything that has a machine channel (impls `MachineWithChannel`) takes care of our usual communication, reducing boilerplate code by a lot. Implementing a `MachineWithChannel` (with no hardware being controlled) takes only around 50 lines of code to send something to the front-end. A big improvement compared to our 4 files in the past.


MilanVDB:
I want to make a Video about this Week 1 January, please finish up the review.

Oshgnacknak:
If my refactoring ethercat connections, that currently goes alongside is undesired/needs more work, please tell me sooner rather then later. For now, I can still split my commits into separate pulls and here, solely focus on getting modbus tcp to work

MilanVDB:
yes, please splitcommits into separate pulls and here, solely focus on getting modbus tcp to work. 

For your refactoring of EtherCAT connections, open a new issue and go ahead

Oshgnacknak:
The non-breaking changes have been distilled out. Now only the new machine gets added.

control-core/src/modbus/tcp.rs
```diff
@@ -0,0 +1,274 @@
...
+            self.read_u16().await?,
+            count as u16,
+            "Modbus device wrote wrong number of registers!"
+        );
+
+        Ok(())
+    }
+
+    pub async fn get_string<const N: usize>(&mut self, addr: u16) -> Result<String> {
+        assert!(N % 2 == 0, "Strings are always of even length!");
```
Snacj, line 193: If an assert! is not fulfilled it will panic. If the panic is not explicitly handled it will crash the program. Is this wanted?

electron/src/control/units.tsx
```diff
@@ -28,10 +28,9 @@ export function getUnitIcon(unit: Unit): IconName {
     case "Hz":
       return "lu:AudioWaveform";
     case "W":
-      return "lu:Zap";
```
Snacj, line 31: Why remove icons for Watt and Volt?

Snacj (dismissed):
See attached comments, otherwise - looks good to me.

electron/src/control/units.tsx
Oshgnacknak: Fall-through casees. 

```
case 1:
  x
case 2:
  x
```
and
```
case 1:
case 2:
  x
```
are semantically equivalent.

control-core/src/modbus/tcp.rs
Oshgnacknak: Yes, because the length is known at compile time by the device documentation. Also there is no way, to read an odd number of because modbus can only respond with 16 bit registers (2 bytes). I wanted to do a static assert, but that need yet another crate. Static asserts are not built in (buh rust!).

As for the other panics, want be to convert them into `Err` values?

PR #925, opened 2025-11-30
891 switch to setup page when selected machine disconnected

EinsPommes:
This fix includes both backend and frontend changes:
- Backend: The serial device detection loop now properly detects disconnections 
  by always checking the device list, even when empty
- Frontend: The UI monitors the machines list and automatically navigates to 
  setup when the active machine disappears

### DIFF
--- a/electron/src/components/SidebarLayout.tsx
+++ b/electron/src/components/SidebarLayout.tsx
@@ -1,7 +1,12 @@
 import { OutsideCorner } from "@/components/OutsideCorner";
 import { useMachines } from "@/client/useMachines";
 import { useOnSubpath } from "@/lib/useOnSubpath";
-import { Link, Outlet, useRouter, useRouterState } from "@tanstack/react-router";
+import {
+  Link,
+  Outlet,
+  useRouter,
+  useRouterState,
+} from "@tanstack/react-router";
 import { Fragment, useEffect } from "react";
 import React from "react";
 import { Icon, IconName } from "./Icon";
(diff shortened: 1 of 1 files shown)


electron/src/machines/laser/laser1/Laser1Graph.tsx
```diff
@@ -19,6 +19,30 @@ export function Laser1GraphsPage() {
...
+      return { value: val.value * 100, timestamp: val.timestamp };
+    };
+    
+    const transformSeries = (series: any) => {
+      if (!series || !series.values) return series;
+      return {
+        ...series,
+        values: series.values.map(transformValue),
+      };
+    };
```
kraemr: Please do not use any here, specify a proper type.

electron/src/components/SidebarLayout.tsx
```diff
@@ -57,7 +58,33 @@ export function useSidebarlessWidth() {
...
+  useEffect(() => {
+    // Wait for machines data to be loaded
+    if (!machinesEvent?.data) return;
+
+    // Extract serial from current path (e.g., /machines/laser1/12345/control)
+    const pathMatch = routerState.location.pathname.match(/\/machines\/[^/]+\/(\d+)/);
+    if (!pathMatch) return; // Not on a machine page
+
+    const serialNumber = parseInt(pathMatch[1], 10);
+
```
kraemr: What if parseInt fails here ? Add error handling and maybe return early if parsing to int fails for whatever reason, also a little pedantic check the size of the pathMatch before accessing at index 1 here

kraemr (changes requested):
Resolve these two issues, then its pretty much mergeable. Btw we usually do one issue --> One pull request not two issues -> One Pull. For smaller tickets this is fine, but just keep it in mind.

electron/src/machines/laser/laser1/Laser1Graph.tsx
EinsPommes: I'll delete it, it wasn't intentional that it was pulled here.

kraemr:
@EinsPommes Fix the Ci failure

EinsPommes:
The errors have been fixed. Is there anything else that needs to be fixed? (Can I merge it?)

kraemr (approved):

PR #941, opened 2025-12-06
936-documentation for minimal example

JalleliAchraf:


MilanVDB:
@kraemr looks good, pls merge :)

PR #969, opened 2025-12-15
968 add mock machines to pr checks and fix mocks and dev build

TheBest6337:
This PR adds mock-machines to pr checks and fixes the mock-machines and dev build features.


Oshgnacknak:
(In a separate pull) we might want to split the rust builds. Format and mock should have their own task imo.

Issue #939, opened 2025-12-04
Harmonize naming of machines


Snacj:
There are some issues in general with naming of machines. We do have naming "conflicts" e.g. our winder is in directory winder1 but it is the WinderV2.
Same with the Extruder.

Please change the naming so that is aligned. /machines/src/winder2 has WinderV2 Struct. /machines/src/winder3 has WinderV3 Struct. etc.

PR #985, opened 2025-12-17
Harmonize machines naming

ldbach:
Issue: Harmonize naming of machines #939 is finished


machines/src/serial/init.rs
```diff
@@ -130,7 +130,7 @@ pub async fn handle_serial_device_hotplug(
                 }
 
                 let app_state_event = app_state.clone();
-                let main_namespace = &mut app_state_event
+                let main_namespace = &mut app_state_event2
                     .socketio_setup
                     .namespaces
                     .write()
```
kraemr, line 136: ???? Remove this change

Issue #916, opened 2025-11-23
When assigning the ethercat terminals no keyboard is showing up


TheBest6337:
When an input field is shown, there should always be an keyboard displayed for input.

MilanVDB:
pls review @Oshgnacknak

PR #1000, opened 2025-12-20
995 no keyboard shows up when creating a new preset

EinsPommes:
When creating a new preset, no keyboard or numpad was shown when focusing the preset name input field. This was especially problematic on touch devices.

This change fixes the same issue that was addressed in #916 (EtherCAT terminal assignment) by integrating the `TouchNumpad` into the `NewPresetDialog`. The implementation follows the same pattern used in `DeviceEepromDialog.tsx` to ensure consistent behavior.

The numpad now appears automatically when the input field is focused, and interacting with the numpad no longer causes the dialog to close unintentionally. Input handling now behaves identically to the EtherCAT assignment dialog.

### DIFF
--- a/ethercat-hal/src/devices/mod.rs
+++ b/ethercat-hal/src/devices/mod.rs
@@ -52,9 +52,9 @@ use el7031_0030::EL7031_0030_IDENTITY_A;
 use el7041_0052::EL7041_0052_IDENTITY_A;
 use ethercrab::{MainDevice, SubDeviceIdentity};
 use smol::lock::RwLock;
-use wago_modules::ip20_ec_di8_do8::{IP20_EC_DI8_DO8_IDENTITY, IP20EcDi8Do8};
 use std::{any::Any, fmt::Debug, sync::Arc};
 use wago_750_354::{WAGO_750_354_IDENTITY_A, Wago750_354};
+use wago_modules::ip20_ec_di8_do8::{IP20_EC_DI8_DO8_IDENTITY, IP20EcDi8Do8};
 
 #[derive(Debug, Clone, Copy)]
 pub struct Module {
--- a/ethercat-hal/src/devices/wago_modules/ip20_ec_di8_do8.rs
+++ b/ethercat-hal/src/devices/wago_modules/ip20_ec_di8_do8.rs
@@ -1,8 +1,8 @@
+use super::*;
 use crate::devices::{
     EthercatDevice, EthercatDeviceProcessing, EthercatDeviceUsed, NewEthercatDevice,
     SubDeviceIdentityTuple,
 };
-use super::*;
 use crate::{
     devices::{DynamicEthercatDevice, Module},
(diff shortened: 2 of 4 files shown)


Oshgnacknak:
Stop! This is the wrong approach! Its fine to show a numpad where only numbers are used, as a better alternative to the native on-screen keyboard. We need to really focus on fixing the on-screen keyboard here. This is a nixos issue, that cannot be solved via front-end code alone.

EinsPommes:
Okay, understood. I implemented the numpad specifically as a workaround for the on-screen keyboard issue. Would it be acceptable to keep this solution for now until the NixOS keyboard problem is properly addressed? @Oshgnacknak

Oshgnacknak:
No, we should focus our resources on getting the on-screen keyboard. Feel free to experiment with that instead

EinsPommes:
Okay, I can try again. I’ve already experimented with this multiple times, but without success so far.
Do we know for sure that this is a NixOS-specific issue?
And just to confirm: we’re using the GNOME on-screen keyboard, correct? @Oshgnacknak

Oshgnacknak:
I also did some research again yesterday and I could not even find a consistent way to spawn the keyboard from the command line. The suggested solution is to ask the user to gesture the keyboard by swiping to monitor from the bottom to the top. 

It was then, when I stumbled into https://github.com/DigiThinkIT/electron-virtual-keyboard, which might be a valid alternative. We could use our own on-screen keyboard, that is only used by our electron app. Its not a beautiful solution, but should do the job, better then now. 

The numpad is not a permanent solution here because presets should be named something like "Black 1,75mm PETG". I leave the judgment up to you we can:
- Use the numpad now and move this issue back into backlog. Here, please refactor the code a copilot already suggested into a `useNumpad` hook and maybe a entire on-screen numpad component.
- Instead of the numpad, show the virtual keyboard, again, please create a reusable component.  
- Go further into trying to fix this with nixos.

EinsPommes:
## Proposal: In-App Virtual Keyboard

I suggest we implement an in-app virtual keyboard as the primary fix.

On NixOS, there is currently no reliable or reproducible way to programmatically trigger the system on-screen keyboard. CLI-based approaches are inconsistent, and the “swipe up” gesture is not a viable solution.

The numpad can remain as a component for numeric-only inputs, but it is not sufficient for naming presets (e.g. “Black 1,75mm PETG”).

### Proposed Plan

- Add a reusable `VirtualKeyboard` component (and `useVirtualKeyboard` hook) that attaches to focused text inputs inside the Electron app.
- Refactor the existing numpad into `useNumpad` + `<Numpad />`, limited to numeric fields only.
- Move the NixOS on-screen keyboard investigation into a separate backlog task (Wayland/X11 + desktop environment + minimal reproduction), so we are not blocked by OS-level behavior.
- I would open a separate issue for implementing the in-app virtual keyboard to keep this scoped and trackable.

PR #1017, opened 2025-12-29
Fix nixos electron build

Oshgnacknak:
To fix the build, the docs there was made available during the election.nix build.

Also, the vite and electron-forge configs have been simplified.


TheBest6337 (changes requested):
Image

The install process is still not working

PR #1039, opened 2026-01-09
Fix modbus discovery

Oshgnacknak:
Seems like we both (me an @kraemr) missed it: I had it not search all addresses because this was to slow until I fixed the timeout. This change now checks all addresses of each interface - does not skip 200.


Snacj (approved):
Looks good to me.

PR #1048, opened 2026-01-14
Upgrade to zod v4

Oshgnacknak:
To make this work, I lot if typescript types had to be fixed.
In general, they became a little simpler, so the update is worth it.
Also removes zod-adapter and upgrades zod-validation-error to work with zod v4.

subserves #1026


electron/src/control/EditValue.tsx
```diff
@@ -36,7 +36,6 @@ type Props = {
   maxSlider?: number; // Override the slider max value
   maxLabel?: string;
   step?: number;
-  valueSchema?: z.ZodType<number>;
```
Oshgnacknak, line 39: This option was never used btw

PR #1050, opened 2026-01-17
1003 add a ruff time estimate for the overall progress

EinsPommes:
Bild 26 12 25 um 13 07

### DIFF
--- a/.github/workflows/issue-status-check.yml
+++ b/.github/workflows/issue-status-check.yml
@@ -1,0 +1,213 @@
+name: Stale Assigned Issue Check
+
+on:
+  schedule:
+    # Run every Sunday at 9:00 AM UTC
+    - cron: '0 9 * * 0'
+  workflow_dispatch: # Allow manual triggering for testing
+
+jobs:
+  check-stale-issues:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      issues: write
+      pull-requests: read
+    
+    steps:
+      - name: Check assigned issues for status updates
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const ONE_WEEK_MS = 7 * 24 * 60 * 60 * 1000;
+            const now = new Date();
+            
+            // Helper function to check rate limits
+            async function checkRateLimit() {
+              try {
+                const { data: rateLimit } = await github.rest.rateLimit.get();
+                const remaining = rateLimit.resources.core.remaining;
+                const reset = new Date(rateLimit.resources.core.reset * 1000);
+                console.log(`Rate limit: ${remaining} requests remaining (resets at ${reset.toISOString()})`);
+                return remaining > 10; // Stop if fewer than 10 requests remaining
+              } catch (error) {
+                console.log(`Could not check rate limit: ${error.message}`);
+                return true; // Continue on error
+              }
+            }
+            
+            // Get all open issues
+            const issues = await github.paginate(github.rest.issues.listForRepo, {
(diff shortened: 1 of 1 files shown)


EinsPommes:
Could you please do a review? @Oshgnacknak

Oshgnacknak:
I like it. In fact, configure the alert to be a warning that cannot be closed and states "Update in process, machine will reboot. Don't use in Production until finished", or something similar.

TheBest6337:
I would agree with @Oshgnacknak, it should also be somewhere more in the middle, because not everyone will look at the bottom right corner. I think it could fit somewhere above the percantage value, at the same height as the buttons.

EinsPommes:
Yes, I can do that, but currently the message disappears after a few seconds. So should it be displayed permanently?
@Oshgnacknak @TheBest6337

TheBest6337:
Yes @EinsPommes

EinsPommes:
Bild 25 01 26 um 17 37


That's how it looks. What do you think? @Oshgnacknak @TheBest6337

Oshgnacknak:
@EinsPommes make the message permanent. You can take inspiration from the lasers warning about the diameter not being in spec.

EinsPommes:
Okay, what about the update information, because it actually says the same thing there. @Oshgnacknak

EinsPommes:
Bild 25 01 26 um 19 47
I have now done it this way

TheBest6337:
You are telling the user the same thing in 2 seperate tiles. First it tells you the estimate on the right and then also in the bug update info tile.

EinsPommes:
@TheBest6337 I think everything should be fine for the merge now, right?

TheBest6337:
@JalleliAchraf please review it and also check if everything makes sense

JalleliAchraf:
The only comment I have is that saying that the updates 'typically take 5 minutes' might be an overestimation in the average case. For example it takes far less for already preinstalled versions. I think a better phrasing would be 'updates may take up to 5 minutes'.

electron/src/setup/UpdateExecutePage.tsx
```diff
@@ -108,6 +109,14 @@ export function UpdateExecutePage() {
             Cancel Update
           </TouchButton>
         )}
+        {isUpdating && (
+          <div className="ml-auto flex items-center gap-2 rounded-xl border border-blue-400 bg-blue-600 px-4 py-2.5 text-white shadow-xl backdrop-blur-sm transition-all duration-300">
+            <Icon name="lu:Info" className="h-5 w-5 text-blue-100" />
+            <span className="text-base leading-snug text-blue-50">
+              Updates typically take approximately <strong>5 minutes</strong>
```
JalleliAchraf, line 116: a better phrasing would be 'Updates may take up to 5 minutes' since it isn't really the average case.

TheBest6337:
> The only comment I have is that saying that the updates 'typically take 5 minutes' might be an overestimation in the average case. For example it takes far less for already preinstalled versions. I think a better phrasing would be 'updates may take up to 5 minutes'.

Yes but they can also take longer when lots of rust and electron code has changed

JalleliAchraf:
> > The only comment I have is that saying that the updates 'typically take 5 minutes' might be an overestimation in the average case. For example it takes far less for already preinstalled versions. I think a better phrasing would be 'updates may take up to 5 minutes'.
> 
> Yes but they can also take longer when lots of rust and electron code has changed

it's just a minor detail overall it LGTM

PR #1064, opened 2026-01-21
1002 data export wrong data types for timestamps

Chitransh31:
Fixed the sheets' names, merged the redundant sheets and created a new Auswertung sheet with chart image generation using uPlot. Scatter chart with ExcelJs did not work because of limited chart support and xlsx-chart creates an entirely new excel file for chart generation. Fixed colours in the chart and added legends below the chart for the parameters.

### DIFF
--- a/nixos/packages/electron.nix
+++ b/nixos/packages/electron.nix
@@ -4,7 +4,7 @@ buildNpmPackage rec {
   pname = "qitech-control-electron";
   version = "1.0.0";
 
-  srcs = [ ../../electron ../../docs ];
+  srcs = ../../electron;
   sourceRoot = "electron";
 
   ELECTRON_SKIP_BINARY_DOWNLOAD = 1;
@@ -44,7 +44,6 @@ buildNpmPackage rec {
       --enable-gpu \
       --ignore-gpu-blocklist \
       --enable-gpu-rasterization \
-      --disable-software-rasterizer \
       "$@"
     EOF
     chmod +x $out/bin/qitech-control-electron


MilanVDB:
@TheBest6337 pleaase review, and give me an update what you think about styl & ability

TheBest6337:
@Chitransh31 please fix the build issues that come from code formatting

electron/src/components/graph/excelExport.ts
```diff
@@ -92,16 +113,920 @@ export function exportGraphsToExcel(
...
+      const startDate = new Date(startTime);
+      const endDate = new Date(sortedTimestamps[sortedTimestamps.length - 1]);
+
+      const formatDateTime = (date: Date) =>
+        date.toLocaleString("de-DE", {
+          day: "2-digit",
+          month: "2-digit",
+          year: "numeric",
+          hour: "2-digit",
+          minute: "2-digit",
```
TheBest6337: Why does this function exist so often duplicat code is never good

electron/src/components/graph/excelExport.ts
```diff
@@ -92,16 +113,920 @@ export function exportGraphsToExcel(
...
+        // Add chart image to worksheet
+        const chartImageId = excelJSWorkbook.addImage({
+          base64: chartImage,
+          extension: "png",
+        });
+
+        // Find a good position for the image (after the data and metadata)
+        const lastRow = auswertungWorksheet.rowCount;
+
+        auswertungWorksheet.addImage(chartImageId, {
```
TheBest6337: Only use englisch, no german english mix!

electron/src/components/graph/excelExport.ts
```diff
@@ -92,16 +113,920 @@ export function exportGraphsToExcel(
...
   } catch (error) {
     alert(
       `Error exporting data to Excel: ${error instanceof Error ? error.message : "Unknown error"}. Please try again.`,
     );
   }
 }
 
-// Generate statistics sheet for a graph line
+// Color mapping based on data type to match reference chart
+function getSeriesColor(name: string): string {
```
TheBest6337: Why are you not using the data from the machine itself and instead hardcode fixed values?? If we have a new machine or change the naming you dont want to change something here. It has to be more modular and less hardcoded.

electron/src/components/graph/excelExport.ts
```diff
@@ -92,16 +113,920 @@ export function exportGraphsToExcel(
...
+  // Power - various shades matching reference
+  if (lowerName.includes("total") && lowerName.includes("watt")) return "#2c3e50"; // Dark blue/black
+  if (lowerName.includes("front") && lowerName.includes("watt")) return "#e67e22"; // Orange
+  if (lowerName.includes("middle") && lowerName.includes("watt")) return "#d35400"; // Dark orange
+  if (lowerName.includes("back") && lowerName.includes("watt")) return "#f39c12"; // Yellow-orange
+  if (lowerName.includes("nozzle") && lowerName.includes("watt")) return "#27ae60"; // Green
+
+  // Pressure and RPM - blue shades from reference
+  if (lowerName === "bar") return "#2980b9"; // Blue
+  if (lowerName === "rpm" || lowerName.includes("rpm")) return "#16a085"; // Teal
```
TheBest6337: same issue

electron/src/components/graph/excelExport.ts
```diff
@@ -92,16 +113,920 @@ export function exportGraphsToExcel(
...
+    // Return base64 data (remove the data:image/png;base64, prefix)
+    return imageData.split(",")[1];
+  } catch (error) {
+    console.error("Error generating chart image:", error);
+    return null;
+  }
+}
+
+// Create Auswertung (Analysis) sheet with combined data from all sheets
+async function createAuswertungSheet(
```
TheBest6337: German...

electron/src/components/graph/excelExport.ts
```diff
@@ -92,16 +113,920 @@ export function exportGraphsToExcel(
...
+  // Format time range for title
+  const formatDateTime = (date: Date) =>
+    date.toLocaleString("de-DE", {
+      day: "2-digit",
+      month: "2-digit",
+      year: "numeric",
+      hour: "2-digit",
+      minute: "2-digit",
+      second: "2-digit",
+    });
```
TheBest6337: duplicate...

electron/src/components/graph/excelExport.ts
```diff
@@ -92,16 +113,920 @@ export function exportGraphsToExcel(
...
+    "Temp Front",
+    "Temp Middle",
+    "Temp Back",
+    "Temp Nozzle",
+    "Total W",
+    "W Front",
+    "W Middle",
+    "W Back",
+    "W Nozzle",
+  ];
```
TheBest6337: Why do we need this? This seems unnesecarry complicated

electron/src/components/graph/excelExport.ts
```diff
@@ -92,16 +113,920 @@ export function exportGraphsToExcel(
...
+
+  // Add metadata section after data
+  sheetData.push(Array(columns.length).fill("")); // Empty row
+  sheetData.push(Array(columns.length).fill("")); // Empty row
+
+  // Software information
+  sheetData.push(["Software Information", "", "", "", "", "", ""]);
+  sheetData.push(["Software", "QiTech Control", "", "", "", "", ""]);
+  sheetData.push(["Version", "1.0.0", "", "", "", "", ""]);
+  sheetData.push([
```
TheBest6337: Why tf is this hardcoded??? The Version should never be hardcoded you can get the current installed version in a similar way the ChooseVersionPage.tsx does it.

electron/src/components/graph/excelExport.ts
```diff
@@ -92,16 +113,920 @@ export function exportGraphsToExcel(
...
+    }
+  }
+  // For laser measurement graphs
+  else if (["Diameter", "X-Diameter", "Y-Diameter", "Roundness"].includes(seriesTitle)) {
+    sheetName = seriesTitle;
+  }
+  // Default fallback
+  else {
+    sheetName = seriesTitle;
+  }
```
TheBest6337: Not a fan of having this harcoded here

electron/src/components/graph/excelExport.ts
```diff
@@ -92,16 +113,920 @@ export function exportGraphsToExcel(
...
+  statsRows.push([
+    "Generated",
+    new Date().toLocaleString("de-DE", {
+      year: "numeric",
+      month: "2-digit",
+      day: "2-digit",
+      hour: "2-digit",
+      minute: "2-digit",
+      second: "2-digit",
+    }),
```
TheBest6337: duplicate

electron/src/components/graph/excelExport.ts
```diff
@@ -92,16 +113,920 @@ export function exportGraphsToExcel(
...
+      "Time Range End",
+      lastDate.toLocaleString("de-DE", {
+        year: "numeric",
+        month: "2-digit",
+        day: "2-digit",
+        hour: "2-digit",
+        minute: "2-digit",
+        second: "2-digit",
+      }),
+    ]);
```
TheBest6337: again

electron/src/components/graph/excelExport.ts
```diff
@@ -92,16 +113,920 @@ export function exportGraphsToExcel(
...
+      // Format timestamp as dd.mm.yyyy hh:mm:ss
+      const date = new Date(timestamp);
+      const formattedDate = date.toLocaleString("de-DE", {
+        year: "numeric",
+        month: "2-digit",
+        day: "2-digit",
+        hour: "2-digit",
+        minute: "2-digit",
+        second: "2-digit",
+      });
```
TheBest6337: dupe

TheBest6337 (changes requested):
Im not a fan of this implementation. Its too much code that is A not necesarry or an duplicate of something which could have been done only once. B - too much hardcoded stuff... like the Control Version or all the if cases of string for the machines that can easily break because no one expects this stuff in an excel export. 

You should use more object based programming, I think that it should be possible to remove most of the hardcoded stuff and get the data from the current graph implementation, which would be way cleaner. Some functions also seem to be pretty long and could be made into their own file, this would make an cleaner implementation and you need less duplicate code. If you need to create small changes to an machine file because of that you can do so, the namespaces are there for a reason or you can get the data from the graph page...

Copilots suggestions are also somewhat true..

TheBest6337:
@Chitransh31 whats your status on this, can I review it?

Chitransh31:
> @Chitransh31 whats your status on this, can I review it?

Yes, you can review it.

Chitransh31:
> Code wise its good, i would put most of the files in an extra folder under graph (excel) for example or excelExport. So you can get an better overview.
> 
> When i opened the saved excel file for the Mock-Machine im greeted with an error message:
> 
> Image Image
> please fix that

Screenshot 2026-01-30 at 2 08 40 PM
Refactored to move the excel export files in graph/excelExport folder. I'm able to download and open all the graphs. Please verify.

Chitransh31:
@TheBest6337 please review the code.

TheBest6337:
@Chitransh31 looks good to me, I will still whait with the merge because I want to test it with a larger set of data during production.

TheBest6337 (changes requested):
I just found a bug: When you export once it saves these values but if you then whait a few mins and want to save another set it still only saves to old one. Its possible that you might need to clear some things.

Image
(This was exported at 14:09)

Seems only to happen when staying on the graphs page, if I move to another and then switch back it works as expected.

Image

TheBest6337:
@JalleliAchraf please fix this bug and test in on different machines

TheBest6337:
@JalleliAchraf please work on this

JalleliAchraf:
Will get it done asap

TheBest6337:
@Snacj please review

TheBest6337:
@Snacj ?

PR #1065, opened 2026-01-22
1004 Add historical target line tracking for graphs

### LINKED ISSUE #1004: Target Line with current target
The target line is always set as one horizontal - so in the graph you do not know which target was set before (one can only estimate by the real temperature)

It would be better to update the target line only from the time it was set.

In the image you see how it looks and how it should look in red: 

Image

Change this for every target line not only the pressure one.

Chitransh31:
Added historical tracking for graph “target” lines so targets change only when they were actually updated. If no historical target exists before a timestamp, it falls back to the existing constant value. Winder2 and Laser1 target lines are not updated yet (same pattern can be applied); Mock1 remains static.


MilanVDB:
@TheBest6337 pleaase review, and give me an update what you think about styl & ability

TheBest6337:
@Chitransh31 please fix the code formatting 

```
npm run format -- --write
```

TheBest6337 (changes requested):
Fix the wiggle of the line
https://github.com/user-attachments/assets/abc76004-4cdb-44a4-8138-866b0659cf24

Copilot is correct here, the target lines shouldnt update constantly to the timeseries. This uses too much RAM and has lots of data were not using. Create an better implementation that uses the data from the changing event and just saves this with an timestamp, which is more efficient.

Chitransh31:
Worked on the comments provided. Please check.

TheBest6337:
https://github.com/user-attachments/assets/ecdcdc7a-4401-4072-a17b-4cecfdbe16e4

if you look at the lines they still are wiggling

Chitransh31:
> Bildschirmaufnahme.2026-01-30.um.11.04.22.mov 
> if you look at the lines they still are wiggling

I tried to eliminate the wiggle by following the approach that the target lines once rendered remain static as time goes on and are re-rendered only when the target values change. However, I was only able to eliminate the horizontal wiggle and minimize the slanted/vertical wiggle to some extent. I tried another approach by replicating the way the current parameter lines change on value updates since they don't have any wiggles. However, this approach caused undesirable behaviour of the target lines and caused even more wiggles. The first approach with the minimal slanted/vertical wiggle remains in the latest commit. Please review and verify.

TheBest6337:
its still not very good for me, 1 - 5 min the wiggle is acceptable but when I set the timeframe to 30min its pretty noticable.

TheBest6337:
Closes #1004

TheBest6337:
@JalleliAchraf please finish this

MilanVDB:
@JalleliAchraf please prioritize this and finish this (more important than the WAGO dokumentation right now)

We need this done...

TheBest6337 (changes requested):
~~Youre implementation isnt very RAM efficient, the current implementation stores the current target state every StateEvent like other values are stored. This leeds to lots of data that isnt necesary. Target lines should only store Data to the graphs when the value gets updated.~~

Alina is going to test it our tomorrow and then well see if this need updating.

Rest looks good and I like that it doesnt really move around anymore.

PR #1068, opened 2026-01-22
Fixed fullscreen not working on user machines

Bra7wurs7:
Preload didn't load correctly due to requiring an absolute path.

### DIFF
--- a/electron/src/main.ts
+++ b/electron/src/main.ts
@@ -30,7 +30,7 @@ if (!gotTheLock) {
 }
 
 function createWindow() {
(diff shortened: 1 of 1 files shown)


Oshgnacknak:
The path is absolute due to the wrapper script created in the nixos build. However, this does not seem to help in case of the icon. Fullscreen seems to work fine.

For now, I see no reason to merge this as is, but feel free to investigate further.

TheBest6337:
@Oshgnacknak it does not work when using an PC, no matter the OS (on our machines its fine)

PR #1086, opened 2026-01-30
Update heater power to the real values for extruder v3

TheBest6337:
Updated to reflect the real power of the heaters


TheBest6337:
@Oshgnacknak quick review please :)

PR #1087, opened 2026-01-30
fixed Wago Initialization Logic

### LINKED ISSUE #1049: Wago Bit Offset Mismatch
There seems to be an issue for some Wago Terminals in the init where they receive the offset of the following Terminal. While its being rewritten, add automatic detection of has_rx has_tx by parsing the IOM (Module Id). The Module ID tells us if there is rx tx channels etc

kraemr:
fix #1049 
Now devices start of inputs/outputs should be aligned correctly


Snacj (approved):
Looks good to me.

PR #1114, opened 2026-02-04
feat: automate npm deps hash update for Electron

TheBest6337:


TheBest6337:
So the npmDepsHash will now get updated automatically if an change has been detected. Then the user has to manually start the ci checks again (because github doesnt allow actions to trigger them). The Action will send a message with an explanation on how to do it.

PR #1118, opened 2026-02-04
889 dre keep zoom level when switich from live view to historical view

Oshgnacknak:
Seems to work fine. I can set 1min for live. Select a region; and when going back to live, Im also back at 1min.

### DIFF
--- a/electron/src/components/graph/historicalMode.ts
+++ b/electron/src/components/graph/historicalMode.ts
@@ -3,7 +3,12 @@ import { useRef, useCallback } from "react";
 import uPlot from "uplot";
 import { seriesToUPlotData } from "@/lib/timeseries";
 import { getPrimarySeriesData, stopAnimations } from "./animation";
-import { BigGraphProps, HistoricalModeHandlers, AnimationRefs } from "./types";
+import {
+  BigGraphProps,
+  HistoricalModeHandlers,
+  AnimationRefs,
+  SwitchOrigin,
+} from "./types";
 
 export function useHistoricalMode({
   newData,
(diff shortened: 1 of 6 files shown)


electron/src/components/graph/types.ts
```diff
@@ -3,6 +3,8 @@ import { Unit } from "@/control/units";
 import { TimeSeries } from "@/lib/timeseries";
 import { RefObject } from "react";
 
+export type SwitchOrigin = "button" | "gesture" | null;
```
Oshgnacknak: The type should not be nullable. Instead, use `?` in a parameter, where appropriate.

electron/src/components/graph/historicalMode.ts
```diff
@@ -120,12 +120,43 @@ export function useHistoricalMode({
...
 
   // Switch to historical mode
-  const switchToHistoricalMode = useCallback(() => {
-    isInHistoricalModeRef.current = true;
-    captureHistoricalFreezeTimestamp();
-    stopAnimations(animationRefs);
-    lastProcessedCountRef.current = 0; // Reset processed count
-  }, [captureHistoricalFreezeTimestamp, animationRefs, lastProcessedCountRef]);
+  const switchToHistoricalMode = useCallback(
+    (origin?: "button" | "gesture" | null) => {
```
Oshgnacknak: Use the type instead?

electron/src/components/graph/useGraphSync.ts
```diff
@@ -19,6 +19,7 @@ export function useGraphSync(exportGroupId?: string) {
     null,
   );
 
+  const historicalSwitchOriginRef = useRef<"button" | "gesture" | null>(null);
```
Oshgnacknak: Again, use the type instead of inlineing it

PR #1124, opened 2026-02-05
Don't store entire systemd log in electron.

Oshgnacknak:
Instead, query only the lines that need to be rendered.


kraemr:
We now dont store logs in the frontend at all, closing

PR #1139, opened 2026-02-13
Prevent TypeError in FullscreenButton when running on non-NixOS environments

### LINKED ISSUE #1067: Fullscreen button not working on some machines
Image

Pizzalord8345:
Description:
- This PR addresses the TypeError: Cannot read properties of undefined (reading 'fullscreen') reported in issue #1067.

The Problem:
- On standard PC environments (non-NixOS), the preload.js might not load correctly due to path differences, causing window.electronWindow.fullscreen to be undefined.

The Solution:
- Instead of changing the path logic in main.ts (which was previously discussed and declined in #1068), I added a defensive check using optional chaining in FullscreenButton.tsx.

This ensures that:

    The application remains stable and does not crash on dev-PCs.

    No changes to the existing NixOS absolute path wrapper logic are required.

Closes #1067.

### DIFF
--- a/electron/src/components/FullscreenButton.tsx
+++ b/electron/src/components/FullscreenButton.tsx
@@ -20,24 +20,24 @@ export function FullscreenButton() {
 
   return (
     <Button
-        onClick={() => {
-            if (window.electronWindow?.fullscreen) {
-                if (isFullscreen) {
-                    window.electronWindow.fullscreen(false);
-                    setIsFullscreen(false);
-                } else {
-                    window.electronWindow.fullscreen(true);
-                    setIsFullscreen(true);
-                }
-            } else {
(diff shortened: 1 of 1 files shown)


Snacj:
@Bra7wurs7 since you encountered the error on one of you machines (?) could you test if this solves the issue?

TheBest6337:
@Snacj doesnt solve the issue just creates an error message telling the user that it doesnt work

Pizzalord8345:
@Oshgnacknak I've been investigating the root cause of this issue further. I believe the problem lies in how the preload path is resolved on non-NixOS systems. Would changes to main.ts be acceptable, as long as the NixOS build behavior remains unaffected? I have a potential fix that works locally on Windows/WSL and uses a fallback path resolution without touching any NixOS scripts.

PR #1144, opened 2026-02-19
1185 target line historical

### LINKED ISSUE #1004: Target Line with current target

JalleliAchraf:
-made target values a step function of time(traceable history)
-set dashed line speed fast enough to not notice any more jittery motion


Snacj (approved):
Looks good to me.

PR #1149, opened 2026-02-24
1148 bug troubleshoot page crashes electron

### LINKED ISSUE #1148: BUG:  Troubleshoot page crashes Electron
The Troubleshoot page caues memory-pressure through requesting journald constantly, this either only kills electron or both electron and journald.

**To Reproduce**
Run the Software for a long time -> ~3 Days -> Crash
Or Spam the Frontend with Print statements -> Crash

It Crashes due to way to frequent requests to journald

**Expected behavior**
It should not crash obviously.

**Screenshots/Photos**
If applicable, add screenshots from the HMI display or photos of the machine screen to help explain your problem.

Not sure when this regression happened ...

Solution:
Simply get rid of the bloated and frankly useless terminal wrapper for electron.
Instead make an export button that writes the journalctl -xb into a file with the current date + time of creation as filename.

We dont have time to spend on stuff that is aesthetic at best ...

kraemr:
fix #1148


just-some-entity (approved):
Looks fine to me

PR #1150, opened 2026-02-24
861 winder laser connection

just-some-entity:
Adds a simple way for machines to send connected machines their live values periodically via MachineMessages and implement the capability for the winder to receive live values from the laser machine.


TheBest6337:
@just-some-entity please fix the build issues, also has this been tested between the machines?

PR #1151, opened 2026-02-24
Updated documentation for rest-api

kraemr:


kraemr:
It definitely works @Snacj but id like your input on the documentation, do you think its clear enough?

docs/rest-api.md
```diff
@@ -1,26 +1,71 @@
...
+# username=machine
+# password=VoMZKZK4K4ZIvON+OU2ZcCXy
+# With -k you can ignore curl's self signed certificate error
+curl https://192.168.130.195/api/v2/machine -k -u machine:VoMZKZK4K4ZIvON+OU2ZcCXy
+```
+
+For HTTPS+Authentication we have Caddy (a reverse proxy, with automatic https) in front of the internal API.
+In theory the HTTPS-Api Can be accessed over any network interface that is configured.
+
+If DNS is available on the subnet, you may be able to resolve `qitech.control`.
```
Oshgnacknak: We should just remove from the documentation this. Unless DHPC is handled by the panel, its wrong anyway.

docs/rest-api.md
```diff
@@ -1,26 +1,71 @@
...
+These are untrusted by default in browsers,tools like curl and https libraries.
+
+For most programs you can either ignore this warning or you can manually trust the certificate on your system.
+How these are installed depends on your system.
+
+On Nixos the root Certificate for Caddy is located at: /var/lib/caddy/.local/share/caddy/pki/authorities/local/root.crt
+You just need to copy this file to an usb drive or copy it some other way and add it as a trusted root certificate in your OS.
+
+How this is done heavily depends on your OS and is not further expanded on.
+A single Google seach for your OS should give you the answer.
```
Oshgnacknak: I don't like this paragraph. Firstly, we should not advocate for google (or any other service). Just call it "search on the internet". Also it sounds a bit informal, if not even passive aggressive to me. 
Try:
> Please refer to your OS' documentation or search the Internet about how to install a trusted SSL certificate.

docs/rest-api.md
```diff
@@ -1,26 +1,71 @@
...
+
+
+## Bypassing HTTPS and Authentication
```
Oshgnacknak, line 49: In the future, we should allow setting these by create a `/etc/qitech.nix` that controls firewall and is conditionally loaded during install.

(This will prob. go into its own issue, but I stated it here, because it feels relevant)

docs/rest-api.md
kraemr: Oh yeah it does sound a bit passive aggressive XD

PR #1159, opened 2026-02-25
Fix workflow permission issues

TheBest6337:
Adds the permissions that are needed, so they now only can read → fixes the security issue


TheBest6337:
@Oshgnacknak or @kraemr please check

PR #1169, opened 2026-02-26
1190 target line revisited

JalleliAchraf:
- historical behavior now persists across all time views


electron/src/components/graph/TargetDashOverlay.tsx
```diff
@@ -18,61 +19,12 @@ type ClipRect = {
...
-      hi = mid - 1;
-    }
-  }
-
-  return ans;
-}
-
-function buildSteppedPath(
+function buildSteppedPathFull(
   u: uPlot,
```
Oshgnacknak: I don't know what a uPlot is. For now, pls call this argument `uPlot` or `plot` or whatever.

electron/src/components/graph/TargetDashOverlay.tsx
```diff
@@ -18,61 +19,12 @@ type ClipRect = {
...
-    } else {
-      hi = mid - 1;
-    }
-  }
-
-  return ans;
-}
-
-function buildSteppedPath(
+function buildSteppedPathFull(
```
Oshgnacknak: Is this/should this be called `DashedLine`?

Oshgnacknak (approved):
I think we can merge, als long as this works on the machine and "looks good"

PR #1201, opened 2026-03-03
861 winder laser connection

### LINKED ISSUE #861: Winder Laser Connection
Implement the Connection between the Laser and the Winder.

just-some-entity:

Adds the capability for a machine to subscribe to another machine and receive it's live values as well as state data.
This is achieved via a machine manager handling machines and their subscriptions with minimal overhead.

Using this new system, the winder2 can now seamlessly subscribe to the laser to receive it's data whenever it is modified.

Additionally, removed all MachineMessage::ConnectToMachine and MachineMessage::DisconnectMachine since it is now superseded by the new system and wasn't utilized by any machine either way.

## Screenshots
image

- [*] Tested locally
- [*] Tested on the machine (if hardware interaction is involved)
- [*] No format errors (`cargo fmt` / `npm run format`)
- [*] No build errors (`cargo build` / `npm run build`)

### DIFF
--- a/electron/src/assets/markdown/winder/manual.md
+++ b/electron/src/assets/markdown/winder/manual.md
@@ -518,6 +518,59 @@ All settings such as the position of the spool sides, the winding speed or the f
 
 <br><br>
 
+## 2.3.1 Puller Adaptive Mode
+
+The Puller Adaptive Mode automatically adjusts the puller speed based on live diameter measurements from a connected laser measurement device. This keeps the filament diameter close to the target value without manual speed adjustments.
+
+<br>
+
+**Prerequisites**
+
+Before the adaptive mode can work correctly, two things must be set up:
+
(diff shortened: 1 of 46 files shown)


TheBest6337:
@just-some-entity did you test the adaptive feature? From my overview it looks like it would change the speed constantly which would leed to an uneven diameter --> 1. diameter is to small, then it slows down until it hits the target, but at that point it has slowed down so much that until the diameter has changed it will be way to thic. Also idk why we need deviation limit fixed and relative, they handle the same thing? 

For the extrusion process it just looks to extreme, what will be needed:

- Less extreme change of speed, (even when over or under the diameter limits)
- An algo that takes the extruded distance into account, as the puller speed change only changes the diameter after x meters, this can take some time and at the moment it could adjust the speed, even tough it should whait for the current changes (if they dont take effect after x meters it can adjust even more)
- There could also be a Deadzone between to values where nothing changes, and only if its above the value it changes, so we dont need to constantly adjust the puller speed, even tough its pretty constant with the diameter.

TheBest6337:
I updated the algorithm now, we only need to test it out and find good working default values.

TheBest6337 (approved):

PR #1213, opened 2026-03-06
Added driver and minimal machine for wago 750-553

Bra7wurs7:


TheBest6337:
@Bra7wurs7 i sadly had no time to test it out, please do an format and I will assign someone else to approve the pr

electron/src/machines/minimal_machines/wago750553machine/useWago750_553Machine.ts
```diff
@@ -0,0 +1,96 @@
...
+  const { request: sendMutation } = useMachineMutate(
+    z.object({
+      action: z.string(),
+      value: z.any(),
+    }),
+  );
+
+  const debounceTimers = useRef<Map<string, ReturnType<typeof setTimeout>>>(new Map());
+
+  const debounced = (key: string, fn: () => void, delay = 150) => {
```
Oshgnacknak: Not sure if we even need debouncing. If so, I would like this in a lib folder called something like `useDebounced`.

electron/src/machines/minimal_machines/wago750553machine/Wago750_553MachineControlPage.tsx
```diff
@@ -0,0 +1,99 @@
...
+                <TouchSlider
+                  value={[value]}
+                  min={0}
+                  max={1}
+                  step={0.001}
+                  minLabel="0 mA"
+                  maxLabel="20 mA"
+                  renderValue={(v) => `${(v * 20).toFixed(2)}`}
+                  unit="mA"
+                  onValueChange={([v]) => {
```
Oshgnacknak: Perhaps its enough to use `onEnter` instead of `onChange`. The former is debounced by the browser. Only emits the event, when the user releases the element.

machines/src/minimal_machines/wago_750_553_machine/act.rs
```diff
@@ -0,0 +1,45 @@
...
+            }
+            MachineMessage::HttpApiJsonRequest(value) => {
+                use crate::MachineApi;
+                let _res = self.api_mutate(value);
+            }
+            MachineMessage::ConnectToMachine(_machine_connection) => {}
+            MachineMessage::DisconnectMachine(_machine_connection) => {}
+            MachineMessage::RequestValues(sender) => {
+                sender
+                    .send_blocking(MachineValues {
```
Oshgnacknak: Can we use `smol::spawn` to send in a task non-blocking instead?

Oshgnacknak (changes requested):

machines/src/minimal_machines/wago_750_553_machine/act.rs
Bra7wurs7: done

electron/src/machines/minimal_machines/wago750553machine/Wago750_553MachineControlPage.tsx
Bra7wurs7: done

electron/src/machines/minimal_machines/wago750553machine/useWago750_553Machine.ts
Bra7wurs7: fixed

Snacj:
Please always provide a description for the Pull Requests with at least:

fix #960

TheBest6337:
@Oshgnacknak please review

PR #1242, opened 2026-03-10
Added driver + minimal machine for 750_460

Bra7wurs7:


TheBest6337:
@Bra7wurs7 the flake.nix issues have nothing to do with the main issue you want to solve with this pr. Please remove it, because https://github.com/qitechgmbh/control/pull/1164 should already fix it.

cargo_run_linux.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
```
TheBest6337: Why is this change needed?

Bra7wurs7: So the script works on nixos which doesnt have bash at /bin/
image

TheBest6337:
@Oshgnacknak please review

cargo_run_linux.sh
Oshgnacknak: That was my suggestion. Other Distros have a symlink by default, but that is not the Nix way I guess

MilanVDB:
@Bra7wurs7 pls fix

cargo_run_linux.sh
ninelore: the correct way to do this is actually 
```suggestion
#!/usr/bin/env bash
```

ninelore: Also FYI POSIX `sh` is s slightly different language than bash. In more complex scripts changing the shebang can have breaking side effects

Bra7wurs7:
Everything fixed, sorry it took so long 🙈
~~edit: actually everything broke, give me a sec~~
edit2: had borked the merge, everything fixed now

cargo_run_linux.sh
Bra7wurs7: implemented suggestion from @ninelore

TheBest6337:
@kraemr please approve if everything looks fine now

Issue #1255, opened 2026-03-16
clean and improve nix code


ninelore:
Currently the nix files violate alot of best practices and have alot of LSP warnings, which can result in degraded build performance and sudden breakages

TheBest6337:
@ninelore please assign yourself when working on an issue, also create the branch by clicking on development -> create branch and then create a branch with the recommended naming scheme.

ninelore:
This issue shouldn't be closed by #1257. That PR only does the minimum to implement the ISO in a clean way and there is still room for improvement, particularly in packaging

PR #1257, opened 2026-03-16
Feat: NixOS custom ISO

ninelore:
CUstom NixOS ISO 9660 Image, some nix fixups necessary for the implementation and some formatting with `nixfmt` RFC style

- [x] Tested locally

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -97,7 +97,9 @@
               nixd
             ];
 
-            ELECTON_SKIP_BINARY_DOWNLOAD = 1;
+            ELECTRON_SKIP_BINARY_DOWNLOAD = 1;
+            # TODO: There's probably a better solution
+            ELECTRON_OVERRIDE_DIST_PATH = "${electron}/bin/";
 
             hardeningDisable = [ "fortify" ];
 


nixos/os/configuration.nix
```diff
@@ -290,45 +277,56 @@ in {
...
-  environment.variables = {
-    QITECH_OS = "true";
-    QITECH_OS_GIT_TIMESTAMP = gitInfo.gitTimestamp;
-    QITECH_OS_GIT_COMMIT = gitInfo.gitCommit;
-    QITECH_OS_GIT_ABBREVIATION = gitInfo.gitAbbreviation;
-    QITECH_OS_GIT_URL = gitInfo.gitUrl;
-  };
+  environment.variables =
+    let
+      gitInfo = import ../gitInfo.nix { inherit pkgs; };
```
Oshgnacknak: you import the gitInfo at two places

nixos/gitInfo.nix
```diff
@@ -1,5 +1,9 @@
 { pkgs, ... }:
 
+## Alternative implementation without using env
+# gitTimestamp="date --date='@${src.lastModified}' --iso-8601=seconds"
+# gitCommit="${src.rev or src.dirtyRev}"
```
Oshgnacknak, line 5: If you can get rid of the (impure) env vars, please do so. We do need all of the vars however and list time I tried to tidy this up, this was the best I could do

flake.nix
```diff
@@ -5,121 +5,133 @@
...
-      inputs.nixpkgs.follows = "nixpkgs";
+    crane = {
+      url = "github:ipetkov/crane";
     };
 
-    # Add the QiTech Control repository
-    qitech-control = {
-      url = "github:qitechgmbh/control";
+    home-manager = {
+      url = "github:nix-community/home-manager/release-25.11";
```
Oshgnacknak, line 13: Is it ok that here, we use 25.11 and in the home.nix, we use 24.xx?

flake.nix
```diff
@@ -5,121 +5,133 @@
...
 
-        # Use Rust 1.86 stable from nixpkgs
-        rust = pkgs.rustc;
-      in {
-        packages = {
-          server = pkgs.qitechPackages.server;
-          electron = pkgs.qitechPackages.electron;
-          default = self.packages.${system}.server;
-        };
+            ELECTON_SKIP_BINARY_DOWNLOAD = 1;
```
Oshgnacknak: I think this should only be in the electron.nix

nixos/gitInfo.nix
ninelore: Its just these 2 (out of 5 or 6) where its possible. I dont think theres a good way around it. I wrote em down for possible future use.
Also we wont be able to drop `--impure` anyway as long as we want to reference `/etc/nixos/hardware-configuration.nix` (or any other file outside the repo

flake.nix
ninelore: tldr its ok, even recommended, to not change it.

Long version: `stateVersion` is a different thing than the actual NixOS Version. Theres a dedicated FAQ for it which should be consulted for more Info: https://wiki.nixos.org/wiki/FAQ/When_do_I_update_stateVersion

flake.nix
ninelore: [This is unchanged from current master](https://github.com/qitechgmbh/control/blob/c5a4abf06dc8d725b54d1a60850c338daa231010/flake.nix#L66). This is probably something for #1255

ninelore:
> we should move our nix files into their own repo?

This repo is effectively a monorepo already since this already houses electron and multiple different rust crates, both library and binary. Changing this would warrant a general policy discussion which is definetly out of scope of this PR.

Regarding the shell scripts: yes, moving them inside `nixos/` would be better, but adapting them is a non-trival task out of scope of this PR

nixos/os/home.nix
```diff
@@ -1,31 +1,40 @@
...
     "${pkgs.qitechPackages.electron}/share/applications/de.qitech.control-electron.desktop";
 
   dconf.settings = {
     # Set GNOME wallpaper
     "org/gnome/desktop/background" = {
-      picture-uri =
-        "https://i.postimg.cc/Z5XJtNMW/qitech-industries-wallpaper.jpg";
-      picture-uri-dark =
-        "https://i.postimg.cc/Z5XJtNMW/qitech-industries-wallpaper.jpg";
+      picture-uri = "https://i.postimg.cc/Z5XJtNMW/qitech-industries-wallpaper.jpg";
```
Bra7wurs7, line 20: We could consider downloading the image and embedding it into the iso. That should work, right?

nixos/os/iso.nix
```diff
@@ -0,0 +1,43 @@
...
+    variables = {
+      # Fix calamares fractional scaling
+      QT_QPA_PLATFORM = "$([[ $XDG_SESSION_TYPE = \"wayland\" ]] && echo \"wayland\")";
+    };
+    defaultPackages = with pkgs; [
+      gparted
+      vim
+      nano
+      # Calamares for graphical installation
+      calamares-nixos
```
Bra7wurs7, line 33: does installation-cd-base even have a windows manager to display calamares?
nixos-setup allows for unattended installation, right? Do we even need calamares?

ninelore: We import `configuration.nix`, which results in the ISO having almost everything set and installed as a Installation. The only difference is the filesystem (squashfs inside ISO 9660) and bootloader (GRUB configured for isos)

nixos/os/home.nix
ninelore: Yes, but i would consider this out of scope for this PR. See also #1255

flake.nix
Bra7wurs7: While you're at it, please also fix
The typo:
          ```ELECTON_SKIP_BINARY_DOWNLOAD = 1;```
to
          ```ELECTRON_SKIP_BINARY_DOWNLOAD = 1;```
and add
          ```ELECTRON_OVERRIDE_DIST_PATH = "${pkgs.electron}/bin/";```
to explicitly tell electron to use the bin from the nix store.
I believe the latter change to be necessary but am not completely sure. I didn't get it to work otherwise. No idea if electron actually is smart enough to use the nix-store binary when it doesnt download it.

Had already started a PR, but to avoid merge conflicts this is probably smartest.

nixos/os/configuration.nix
ninelore: fixed

PR #1262, opened 2026-03-20
flake.nix electron fixes

Bra7wurs7:


TheBest6337:
In what way does this differ from https://github.com/qitechgmbh/control/pull/1257 ?

Bra7wurs7:
> In what way does this differ from #1257 ?

Fixes the "electon" typo and adds a line to explicitly tell electron to use the binary provided in the nix store to fix the missing libs

PR #1294, opened 2026-04-03
Auto fix rust linting issues

JHaertl:

- Fixes typo in lint setup ("pe**n**dantic" in Cargo.toml)
- Adds githooks for formatting and linting
- Fixes all auto fixable clippy issues

This currently does not add linting to CI checks as there are many more linting issues that clippy could not fix.
Just a proposal towards a path without clippy issues to increase code quality.
I appreciate the fact that this comes with tons of changes from a random guy. Happy for you to apply the clippy fixes yourself to save review time.

I am also happy to adapt the current clippy rules upfront as the non-default categories like pedantic and nursery can be controversial latest when it comes to activating this in CI.
Going forward I would propose to:
- fix all default category issues
- mark the default categories (anything that is warn by default as described here https://doc.rust-lang.org/stable/clippy/) as deny, keep the remaining ones as warnings
- run clippy with warnings as errors in the pre-commit hook to encourage dealing with the warnings (hooks can be skipped)
- run clippy with the above defaults in CI

Consider this PR the start of a conversastion.

*Note that all auto-fixes are applied without `--broken-code`, wanted to leave that and further fixes for follow-up PRs.
Once the repo has no clippy issues we can add clippy to the rust CI steps.

### DIFF
--- a/control-core/src/modbus/tcp.rs
+++ b/control-core/src/modbus/tcp.rs
@@ -52,7 +52,7 @@ pub struct ModbusTcpDevice {
 impl Debug for ModbusTcpDevice {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self.stream.peer_addr() {
-            Ok(addr) => write!(f, "ModbusTcpDevice({:?})", addr),
+            Ok(addr) => write!(f, "ModbusTcpDevice({addr:?})"),
             Err(_) => std::fmt::Result::Err(std::fmt::Error),
         }
     }
@@ -98,9 +98,7 @@ impl ModbusTcpDevice {
         if func_code != func_code_res {
             let error_code = self.read_u8().await?;
             bail!(
-                "Modbus device sent unexpected function code 0x{:x}! It probably encountered an error. Error code: 0x{:x}",
-                func_code_res,
-                error_code
+                "Modbus device sent unexpected function code 0x{func_code_res:x}! It probably encountered an error. Error code: 0x{error_code:x}"
             );
         }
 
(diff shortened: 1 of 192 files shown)


TheBest6337:
@Oshgnacknak can you take a look please?

control-core/src/modbus/tcp.rs
```diff
@@ -190,7 +187,7 @@ impl ModbusTcpDevice {
     }
 
     pub async fn get_string<const N: usize>(&mut self, addr: u16) -> Result<String> {
-        assert!(N % 2 == 0, "Strings are always of even length!");
+        assert!(N.is_multiple_of(2), "Strings are always of even length!");
```
Oshgnacknak, line 190: finally this goes through the compiler

server/src/panic.rs
```diff
@@ -22,8 +22,8 @@ fn panic_hook(panic_info: &PanicHookInfo) {
         .unwrap_or_else(|| "<unknown>".to_string());
 
     tracing::error!("thread '{}' panicked at {}:", thread, locataion);
```
Oshgnacknak, line 24: "locataion" we might fix this typo while diffing this file

Oshgnacknak (approved):
whats with alle the `must_use`? In what way does this help?

Oshgnacknak:
i think we can merge this. However, I want another opinion on `must_use`. Maybe @kraemr

JHaertl:
@Oshgnacknak I agree that must_use is not super helpful all over the place. As described I mainly wanted to start a conversation about this. If you want I can first create a PR with improved linting rules. 
Id roughly go with
- Deny all default warnings
- Additionally deny pedantic + cargo
  - Cherry-pick allows for certain rules
- Cherry-pick denies from restriction and nursery

Would you be open to that? Something roughly like https://rtaw.co.uk/posts/clippy-lints/

Oshgnacknak:
I don't think we will merge this in its current state. There sadly just isn't the value in adding all these changes on their own, but also all at once.
We are none the less happy with community contributions, but there needs to be more communication upfront.
Maybe @JHaertl, you can open a discussion and suggest what changes you want to contribute, then we could group then together in a appropriate pull request.

PR #1307, opened 2026-04-19
Added drywell machine and UI

khalilbendhief:
Added the Digicolor Drywell integration


machines/src/serial/devices/drywell/mod.rs
```diff
@@ -0,0 +1,375 @@
...
+            .send(DrywellWriteCommand::WriteCoil {
+                address: 273,
+                value: true,
+            })
+            .await?;
+        self.cmd_sender
+            .send(DrywellWriteCommand::WriteCoil {
+                address: 273,
+                value: false,
+            })
```
kraemr: When setting an address for a request please atleast define a constant instead of just a magic number. I have no idea what address 272 does. Lets say it sets the dew point, then it should be const DEW_POINT_ADDRESS : u16 = 272 for example

machines/src/serial/devices/drywell/mod.rs
```diff
@@ -0,0 +1,375 @@
...
+                            if let Some(&val) = regs.first() {
+                                if val != u16::MAX {
+                                    temp = val as f64;
+                                }
+                            }
+                        }
+                    }
+                }
+                temp
+            };
```
kraemr: Way too much identation here. Limit yourself to like 2-3 levels of identation please.

kraemr (changes requested):

TheBest6337:
This shouldnt be inside the main repo, please move it to: https://github.com/qitechgmbh/digicolor-drywell-control

(You probably also have to update the repo there to match the current master here with ```git push --mirror https://github.com/qitechgmbh/control```)

PR #1350, opened 2026-05-05
Fix machines showing before subdevice initialization

### LINKED ISSUE #1349: BUG: Clicking on Machines before SubDevice init crashes the machine
**Describe the bug**
When the backend process starts, it can happen that the machine shows up before the subdevice initialisation, when you click on the machine and want to put it in hold for example, the machine crashes and becomes unresponsive. The SubDevices do show up after a few seconds, but even when going back you cant control anything and you have to restart the backend process, and whait with opening a machine until the subdevices are initialized.

**To Reproduce**
1. Restart the Backendprocess
2. In the moment when the extruder/winder appears open it
3. Change some things on the machine, e.g. go to hold mode on the winder
4. Nothing happens
5. See that the subdevices have now been initialized
6. Machine is still unresponsive

**Expected behavior**
Machines should only show up after the subdevices are initialized, so you can directly start using the machine.

**Log**
[backend_restart_bug.log](https://github.com/user-attachments/files/27400738/backend_restart_bug.log)

**Machine Information (please complete the following):**
 - Machine Type: Winder2, Extruder (does not happen with the Laser)
 - Control Software Version: Newest Master Commits, 2.15, etc.
 - Settings: doesnt matter

**Device Information (if issue occurs on external device):**
 - Device Type: Control Panel
 - OS: QiTech NixOS Control

TheBest6337:
Fix for: https://github.com/qitechgmbh/control/issues/1349, where an user was able to open the machine, before it was initialized, which lead to being unable to control it and the user having to restart the backendprocess.

Adds an pending machine obj, where all machines are appended until the subdevices are initialized. And only then the machines can show up in the frontend. Flow after the fix:
```
setup_loop → creates machines → stores as pending (NOT visible)
         → EthercatSetup with all_operational=false → RT loop
RT loop  → copy_ethercat_inputs busy-waits for all_op()
         → returns Ok(true) → sends MachinesInitialized via main channel
async handler → flush_pending_machines() → adds to frontend list → MachinesEvent emitted
frontend → receives MachinesEvent → renders machines
```

- [x] Tested locally

### DIFF
--- a/machines/src/lib.rs
+++ b/machines/src/lib.rs
@@ -72,8 +72,6 @@ pub enum AsyncThreadMessage {
     NoMsg,
     SubscribeToMachine(MachineSubscriptionRequest),
     UnsubscribeFromMachine(MachineSubscriptionRequest),
-    /// Sent by the RT loop once all EtherCAT subdevices are operational.
-    /// The async handler should then add pending machines to the frontend list.
     MachinesInitialized,
 }
 
--- a/machines/src/lib.rs
+++ b/machines/src/lib.rs
@@ -72,6 +72,9 @@ pub enum AsyncThreadMessage {
     NoMsg,
     SubscribeToMachine(MachineSubscriptionRequest),
     UnsubscribeFromMachine(MachineSubscriptionRequest),
+    /// Sent by the RT loop once all EtherCAT subdevices are operational.
+    /// The async handler should then add pending machines to the frontend list.
+    MachinesInitialized,
 }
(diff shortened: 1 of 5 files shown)


TheBest6337:
@kraemr something like this will probably also be needed in the lib

ninelore (changes requested):
```sed
s/\/\/\//\/\//
```

ninelore (approved):

PR #1360, opened 2026-05-08
added pre-commit hook + setup script

Bra7wurs7:
fail
succ


ninelore (approved):
IMO a bit overengineered, but lgtm

PR #1370, opened 2026-05-19
added 24h journalctl export instead of since boot,

kraemr:


server/src/main.rs
```diff
@@ -42,7 +42,7 @@ use std::{collections::HashMap, sync::Arc, time::Duration};
 const CYCLE_TARGET_TIME: Duration = Duration::from_micros(2000);
 
 #[cfg(not(feature = "development-build"))]
-const CYCLE_TARGET_TIME: Duration = Duration::from_micros(700);
+const CYCLE_TARGET_TIME: Duration = Duration::from_micros(1000);
```
ninelore, line 45: Is this supposed to be in this PR? Seems unrelated

kraemr: yes milan wanted a quick and dirty "fix", and i dont want to make another pull request for one line of code

PR #1376, opened 2026-05-22
Qitech lib legacy eeprom

Snacj:
This Pull request allows to boot subdevies into preop to write to EEPROM.


Snacj:
@kraemr  Did I break something here with the npm shenanigans?

Snacj:
It won't install on the panel....

PR #1381, opened 2026-05-26
Nodejs version unify

ninelore:

Use the same nodejs major version across the board, package version cleanup 

- [x] Tested locally

### DIFF
--- a/nixos/packages/server.nix
+++ b/nixos/packages/server.nix
@@ -1,6 +1,4 @@
 {
-  lib,
-  pkgs,
   pkg-config,
   libudev-zero,
   libpcap,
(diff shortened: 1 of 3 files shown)


nixos/packages/server.nix
```diff
@@ -10,7 +8,7 @@
 let
   commonArgs = {
     pname = "server";
-    version = "1.0.0";
+    version = (builtins.fromJSON (builtins.readFile ../../electron/package.json)).version;
```
Oshgnacknak, line 15: would it not be more correct to use `qitech_control/Cargo.toml`. On the other hand, they should:tm: always be in sync

Oshgnacknak (approved):

nixos/packages/server.nix
ninelore: root Cartgo.toml has no version, server Cargo.toml is at 0.1.0 since the dawn of time

PR #1385, opened 2026-05-27
Remove dependency on token for retrieving update targets and changelogs

### LINKED ISSUE #1418: Update Tab improvements
- Only show one row of current releases (we dont need to show every version we ever released), older once can be hidden in a foldable element

just-some-entity:

See title ^^^

- [x] Tested locally


MilanVDB:
@ninelore pls review

ninelore (commented):
Image

ninelore:
A button for a manual refetch would be nice as well

ninelore (changes requested):
see above

ninelore (changes requested):
needs to be rebased onto staging as well

ninelore (changes requested):
rebase on staging and fix lint pls

TheBest6337:
Its now a merge into staging and rebased to the newest commit

TheBest6337:
Also fix #1418

ninelore:
> Also fix #1418

GitHub only automatically acts on such comments in the OP btw

ninelore (changes requested):
we're still cloning after hitting `Apply Update`, which shouldnt be necessary if done right

TheBest6337:
@ninelore it shouldnt clone now for 99% of the updates

electron/src/helpers/ipc/update/git-fetch-utils.ts
```diff
@@ -0,0 +1,235 @@
...
+function getDir(owner: string, name: string) {
+  const tmpDir =
+    process.env.TMPDIR || process.env.TMP || process.env.TEMP || "/tmp";
+
+  return `${tmpDir}/${owner}/${name}`;
+}
+
+async function importIfNotExists(owner: string, name: string) {
+  const dir = path.resolve(getDir(owner, name));
+  const url = `https://github.com/${owner}/${name}.git`;
```
ninelore, line 78: We shouldn't assume GitHub, not everyone wants to be on this ...reputable platform

electron/src/helpers/ipc/update/git-fetch-utils.ts
```diff
@@ -0,0 +1,235 @@
...
+            console.error("Failed to remove stale lock file:", err);
+          }
+        }
+      }
+    } catch (err) {
+      console.error("Failed to walk .git directory for lock cleanup:", err);
+    }
+  }
+  walk(path.join(repoPath, ".git"));
+}
```
ninelore: This is super cursed and will sooner or later backfire

electron/src/helpers/ipc/update/git-fetch-utils.ts
```diff
@@ -0,0 +1,235 @@
...
+
+async function fetchWithRetry(repoPath: string): Promise<void> {
+  const maxRetries = 3;
+  for (let attempt = 0; attempt < maxRetries; attempt++) {
+    try {
+      await runGitCommand(["fetch", "--prune", "--force", "origin"], repoPath);
```
ninelore: why `--force`? Possible cause of necessity for manual `.git/` interventions

electron/src/helpers/ipc/update/git-fetch-utils.ts
```diff
@@ -0,0 +1,235 @@
...
+      return;
+    } catch (error: any) {
+      const msg = error?.message || String(error);
+      if (attempt < maxRetries - 1 && /cannot lock ref/i.test(msg)) {
+        removeStaleLockFiles(repoPath);
+        try {
+          await runGitCommand(["pack-refs", "--all"], repoPath);
```
ninelore: - Running this on every fetch is bad-practice and a possible cause of necessity for manual `.git/` interventions
- The repo is by far not big enough (in terms of commits and tags) to make this necesary in the first place

electron/src/setup/ChangelogPage.tsx
```diff
@@ -23,46 +24,46 @@ export function ChangelogPage() {
...
-    headers: {
-      ...(search.githubToken && {
-        Authorization: `token ${search.githubToken}`,
-      }),
-      Accept: "application/vnd.github.v3+json",
-    },
+  const source: GithubSource = {
+    githubRepoOwner: search.githubRepoOwner,
+    githubRepoName: search.githubRepoName,
   };
```
ninelore, line 30: Git Source should be just a valid remote URI as string since not all git forges have the format `domain.tld/owner/repo`, for example gitlab has suborgs/groups

ninelore (commented):
Nit: Symbol namings should be agnostic from github

electron/src/helpers/ipc/update/git-fetch-utils.ts
kraemr: Should be done in a seperate issue

electron/src/setup/ChangelogPage.tsx
kraemr: Should also be in a seperate issue. Something like "make update process github independent"

ninelore (changes requested):
- Still cloning on `Apply Update`
- Nit: Only show branches and commits updated in the last 2 or 3 months, hide everything else just like versions

TheBest6337:
> * Still cloning on `Apply Update`

you mean the text in the Update info?

ninelore:
> > * Still cloning on `Apply Update`
> 
> you mean the text in the Update info?

Update log also clearly shows git clone progress

TheBest6337:
Well then there was an issue detected (and then we clone) I tested it on the HMI and there it wasnt cloning.

ninelore (changes requested):
Further reduce remote git traffic by:
- Applying an update should use the same clone as the one used in this PR (`getDir(owner, name)`) instead of the current hardcoded `$HOME/control`
- Not using `TEMP` or `/tmp`, since this will get cleared on reboot

TheBest6337:
> Further reduce remote git traffic by:
> 
>     * Applying an update should use the same clone as the one used in this PR (`getDir(owner, name)`) instead of the current hardcoded `$HOME/control`
> 
>     * Not using `TEMP` or `/tmp`, since this will get cleared on reboot

Can be done later, which would improve it, but the current implementation should work. We need to merge this today, because we want to release the 3.0.

ninelore (approved):

PR #1419, opened 2026-06-18
Fix staging startup

TheBest6337:

- [ ] Tested locally


electron/src/client/socketioStore.ts
```diff
@@ -312,6 +312,8 @@ export const useSocketioStore = create<SocketioStore>()((set, get) => ({
     socket.on("disconnect", (reason) => {
       socket.disconnect();
       resetStore(set);
+      
+      window.location.reload();
```
TheBest6337, line 316: Just for this commit to work, will be changed in a later PR to not use this workaround

PR #1488, opened 2026-07-03
Disable touch gestures

TheBest6337:
Disables touch gesture weirdness that opened remote desktop for example when using the device with 2 fingers (sometimes also shown up with just one, it was just buggy...)

### DIFF
--- a/nixos/os/configuration.nix
+++ b/nixos/os/configuration.nix
@@ -181,7 +181,7 @@ in
 
   services.libinput.enable = true;
   services.libinput.touchpad.tapping = true;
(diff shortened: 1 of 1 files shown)


ninelore (changes requested):
Are you sure this wont break interaction with GNOME and/or the virtual keyboard?

TheBest6337:
> Are you sure this wont break interaction with GNOME and/or the virtual keyboard?

Yes I tested it and everything I expected to work, was still working.

ninelore (approved):

PR #1538, opened 2026-07-25
Rewinder 

### LINKED ISSUE #1539: Rewinder machine
adds rewinder machine for rewinding filament from source spool to takeup spool while controlling source/takeup tension arms, puller speed, traverse motion, and automatic stop by length.

user can configure line speed, source/takeup spool diameter, tension arm ranges/targets, prepare tolerance, traverse limits/start position/step/padding, spool speed mode, and automatic stop action.

JalleliAchraf:
Rewinder machine implementation


ninelore (dismissed):
There should be no dependabot commits in here. Please rebase again

electron/src/machines/winder/rewinder/useRewinder.ts
```diff
@@ -0,0 +1,581 @@
...
+    traversePosition,
+    pullerSpeed,
+    takeupSpoolRpm,
+    sourceSpoolRpm,
+    takeupTensionArmAngle,
+    sourceTensionArmAngle,
+    rewindProgress,
+    isLoading: false,
+    isDisabled: false,
+    motionStopped,
```
TheBest6337: These shouldnt be hardcoded, because now you can engage with settings even when the machine is currently unresponsive.

Change it to be similar to the other machines. There we were using stateOptimistic i think?

machine_implementations/src/rewinder/emit.rs
```diff
@@ -0,0 +1,953 @@
...
+                self.rewind_control.takeup_command_angular_velocity()
+            } else {
+                let angular_velocity = self.takeup_spool_speed_controller.update_speed(
+                    t,
+                    &self.takeup_tension_arm,
+                    &self.puller_speed_controller,
+                );
+                angular_velocity
+            }
+        } else {
```
TheBest6337: This seems to be dead code? Because for my understanding we always get false here...

machine_implementations/src/rewinder/new.rs
```diff
@@ -0,0 +1,200 @@
...
+
+        {
+            let mut takeup_ref = takeup_spool.0.borrow_mut();
+            (&mut *takeup_ref).write_config(interface.clone(), takeup_spool.1, &el7041_config)?;
+        }
+
+        let (api_sender, api_receiver) = tokio::sync::mpsc::channel(2);
+
+        let mut source_spool_speed_controller =
+            super::SpoolSpeedController::new_with_tension_range_and_response(
```
TheBest6337: Source Spool Speed Controller seems to be constructed once but never gets update_speed called. sync_source_spool_speed seems to compute the RPM from rewind_control and only calls set_speed. So the Source arms of speed for clamping/speed_factor/speed_factor_change could also be dead code?

TheBest6337 (changes requested):
So there is a need for some changes I commented to some code, but there were also other things I found:

- Default start position of the traverse s implemented not as I expected it to:
  1. Of all it should be in the Config/Settings as its nothing that should change to often
  2. There should be 3 Options you can set: Left Side (at the spool flank), right side of the spool and custom where you can set a value yourself. (Most people only need the left or right side toggle.)

-  Why are some settings disabled when im in pull/wind/etc. It should be possible to change the spool traverse flanks and so on even when in wind/pull mode.

- Same with resetting the winding progress, this should always be possible (why is it disabled in standby for example lol)

- For consistency sort the tabs in the top bar as following: 1. Control, 2. Graphs, 3. Settings, 4. Presets

- It seems like theres a lot of dead code/mutations for example it seems like the following have no effect on the machine motion: SetTakeupSpoolRegulationMode, SetTakeupSpoolMinMaxMinSpeed, SetTakeupSpoolMinMaxMaxSpeed, SetTakeupTensionTarget, SetTakeupSpoolAdaptiveRadiusLearningRate, SetTakeupSpoolAdaptiveMaxSpeedMultiplier, SetTakeupSpoolAdaptiveAccelerationFactor, SetTakeupSpoolAdaptiveDeaccelerationUrgencyMultiplier.

- The PR adds SpoolTensionResponse and six new_with_tension_range* constructors across three shared winder2 files (spool_speed_controller.rs, adaptive_spool_speed_controller.rs, minmax_spool_speed_controller.rs).
Source is constructed exactly once, for source_spool_speed_controller (new.rs:134-138) — and that controller's update_speed is never called. sync_source_spool_speed computes RPM from rewind_control or the pull-mode assist and only calls set_speed. So the Source arms of speed_for_clamping / speed_factor / speed_factor_change are dead code.

TheBest6337 (approved):

PR #1540, opened 2026-07-26
Dryer Machine

KALIX-89:

- [ ] Tested locally


qitech_control/src/main.rs
```diff
@@ -120,6 +124,60 @@ fn setup_ethercat(
...
+        let result = tokio::time::timeout(
+            Duration::from_millis(500),
+            ctx.call(Request::ReadInputRegisters(0x0E, 3)),
+        )
+        .await;
+        matches!(
+            result,
+            Ok(Ok(Ok(Response::ReadInputRegisters(ref regs)))) if regs.len() == 3
+        )
+    })
```
TheBest6337: lags when plugging...

machine_implementations/src/dryer/device.rs
```diff
@@ -0,0 +1,616 @@
...
+                    Duration::from_millis(500),
+                    ctx.call(Request::ReadInputRegisters(0x00, 0x21)),
+                )
+                .await;
+                matches!(
+                    result,
+                    Ok(Ok(Ok(Response::ReadInputRegisters(ref regs)))) if regs.len() == 0x21
+                )
+            });
+        if !identity_ok {
```
TheBest6337, line 220: RT block causes massive lags!

machine_implementations/src/dryer/device.rs
```diff
@@ -0,0 +1,616 @@
...
+            Self::timer_entry_regs(&entry).into(),
+        ));
+    }
+
+    pub fn queue_write_new_timer_entry(&mut self, entry: SmartTimerEntry) {
+        self.write_queue.push_back(Request::WriteMultipleRegisters(
+            SMART_REG_TIMER_NEW,
+            Self::timer_entry_regs(&entry).into(),
+        ));
+    }
```
TheBest6337, line 599: It should clamp smart_timer_slots / SMART_TIMER_MAX_SLOTS, because it could create issues when someone sets an value thats out of bounds over the API.

machine_implementations/src/dryer/device.rs
```diff
@@ -0,0 +1,616 @@
...
+                    for i in 0..slots {
+                        let base = (i * SMART_TIMER_ENTRY_REGS) as usize;
+                        if regs.len() < base + 5 {
+                            break;
+                        }
+                        let cycle = regs[base];
+                        let flags = regs[base + 4];
+                        entries.push(SmartTimerEntry {
+                            weekly: (cycle & 0xFF) != 0,
+                            weekday: (((cycle >> 8) & 0xFF) as u8).trailing_zeros() as u8,
```
TheBest6337: returns 8 when the byte is zero. DryerSmartSchedulePage.tsx renders a <select> with options 0–6, so the field shows blank and any other edit on that row round-trips weekday: 8. Decode explicitly and clamp to 0..=6.

electron/src/machines/dryer/dryer_v1/DryerV1ControlPage.tsx
```diff
@@ -0,0 +1,707 @@
...
+        setRemainingSec(remaining);
+        if (remaining === 0 && !autoStopFiredRef.current) {
+          autoStopFiredRef.current = true;
+          timerStartRef.current = null;
+          handleStartStopRef.current(false);
+        }
+      }
+    }, 1000);
+    return () => clearInterval(id);
+  }, []);
```
TheBest6337, line 215: Both the drying-time and scheduled auto-stop only run while that page is mounted. Navigate away, close the window, or drop the socket and the dryer keeps heating. This belongs in act().

qitech_control/src/main.rs
```diff
@@ -120,6 +124,60 @@ fn setup_ethercat(
     Ok(())
 }
 
+const CH340_VID: u16 = 0x1a86;
+const CH340_PID: u16 = 0x7523;
+
```
TheBest6337: With this code we would only accept these two adapters. Did you confirm with the Team that we didnt deploy any other modbus adapters, which wouldnt work now with this change?

Before we accepted any /dev/ttyUSB0/ttyUSB1...

machine_implementations/src/dryer/material_presets.rs
TheBest6337, line 1: Why do we need material_presets.rs and .ts?

electron/src/machines/dryer/materialPresets.ts
TheBest6337, line 1: Why do we need material_presets.rs and .ts?

machine_implementations/src/dryer/api.rs
```diff
@@ -0,0 +1,149 @@
...
+        abbrev: String,
+        throughput_kg_per_h: f64,
+    },
+}
+
+impl MachineApi for DryerMachine {
+    fn api_mutate(&mut self, request_body: Value) -> Result<(), anyhow::Error> {
+        let mutation: Mutation = serde_json::from_value(request_body)?;
+        match mutation {
+            Mutation::SetStartStop(_) => self.set_start_stop(),
```
TheBest6337: SetStartStop(bool) is a toggle, not a setpoint: auto-stop can start the dryer
machine_implementations/src/dryer/api.rs:106 discards the payload (Mutation::SetStartStop(_) => self.set_start_stop()), and queue_set_start_stop (device.rs:467-476) pulses COIL_START_STOP true→false. The device toggles; the boolean is meaningless.

The frontend assumes it is a setpoint. DryerV1ControlPage.tsx:211 and :227 call handleStartStopRef.current(false) from a 1 Hz setInterval when the drying timer or the scheduled stop time elapses.

Failure scenario: operator sets a stop time for later today on the Schedule page and leaves the dryer idle with the Control page open. autoStopFiredRef starts false and is only reset on a not-running → running transition, so at the scheduled time the app fires SetStartStop(false), the backend pulses the coil, and the idle dryer starts heating. The same holds for any UI path that sends false while already stopped. Identical in the Smart page.

Fix: compare against the device's running state (isRunningStatus/status) in set_start_stop and only pulse when the requested state differs or split into explicit Start/Stop mutations. Rename the variant if the pulse semantics stay.

TheBest6337 (changes requested):
Needs changes (in the comments) aswell as structurally: 

The smart dryer is a near copy of the dryer. Only the two schedule pages differ meaningfully. DryerSmartMachine is DryerMachine plus one smart_data field, and DryerDevice already carries the is_smart flag that would drive the difference. Collapsing to one machine + one set of shared React components with a Smart-only section removes roughly 1,700 lines.

machine_implementations/src/dryer/device.rs
KALIX-89: moved identification and Smart probing off the blocking construction path entirely. new() no longer calls rt.block_on() at all, it now spawns the actor immediately and does the identify/variant checks through the same non-blocking request/response cycle everything else already uses. Worst case is now a few seconds of retries with zero thread-blocking, instead of up to 1.5s stalling the real-time loop.

machine_implementations/src/dryer/device.rs
KALIX-89: queue_write_timer_entry/queue_delete_timer_entry both reject (with a warning log, no write sent) if index >= smart_timer_slots, and smart_timer_slots itself is clamped to [SMART_TIMER_MIN_SLOTS, SMART_TIMER_MAX_SLOTS] when read from the device, so an out-of-bounds index from the API can never reach the wire.

machine_implementations/src/dryer/device.rs
KALIX-89: added an explicit day_bits == 0 → weekday 0 case instead of relying on trailing_zeros() (which returned 8 for a zero byte), plus .min(6) as a safety net on both the decode and the write side. weekday can no longer be anything outside 0..=6

electron/src/machines/dryer/dryer_v1/DryerV1ControlPage.tsx
KALIX-89: Moved the countdown into the backend (DryerMachine::remaining_seconds(), computed every tick via act() → emit_live_values()), frontend now just displays remaining_seconds instead of recalculating it.

electron/src/client/useClient.tsx
```diff
@@ -57,7 +57,7 @@ type Client = {
   }) => Promise<MutationResponseSchema>;
 };
 
-const baseUrl = "http://localhost:3001";
+export const baseUrl = "http://localhost:3001";
```
KALIX-89, line 60: this edit was done so the useMaterialPresets.ts can fetch the material list from the backend

TheBest6337 (approved):

PR #1554, opened 2026-07-31
rm mailmap

TheBest6337:
Isnt used anymore and had no usecase in the past.


ninelore (changes requested):
The file is important and is not unused, however contains partiallty garbage

MilanVDB:
raus damit...

PR #1566, opened 2026-08-03
bump packages

TheBest6337:
Bumps electron deps


Oshgnacknak (approved):
Test on machine ofc. Maybe also rebase beforehand

PR #1567, opened 2026-08-03
ts7

TheBest6337:

- [x] Tested locally


electron/package-lock.json
```diff
@@ -87,7 +87,8 @@
         "prettier-plugin-tailwindcss": "^0.6.14",
         "tailwindcss": "^4.3.3",
         "ts-node": "^10.9.2",
-        "typescript": "^6.0.3",
+        "typescript": "npm:@typescript/typescript6@^6.0.3",
+        "typescript-7": "npm:typescript@^7.0.2",
```
ninelore, line 91: Why do we have both now?

TheBest6337: TS 7 is still missing the API parts (standard microslop) and so we need 6 for the api. TS7 would only bring us performance improvements for ts checking but we could also whait for TS 7.1 which will then bring the api back...

ninelore: if ts7 isnt ready and ts6 is supported then lets just not migrate (yet)

TheBest6337: Well technically its finished, but microslop just tought its a good idea to release without an api. Will close this PR for now...

PR #1597, opened 2026-08-17
1470 feat use cpuset instead of isolcpus and fix rt config

### LINKED ISSUE #1470: Feat: use cpuset instead of isolcpus and fix rt config
Currently we use isolcpus to isolate cores. This is highly inflexible as it cant be changed at runtime, when we do updates they take longer due to not being able to use the isolated cores. For that reason we should use cpusets in cgroup. Code will mostly stay the same too. We only need to either persist our cgroup or write it on every boot. 

With cpuset we can get rid of core isolation for updates and turn it on when done.
Updates would be much faster due to having access to two more cores.

To isolate cores 2 and 3 (counting from zero) on a 4-core cpu for example: 
```bash
# AS ROOT
# Only allow housekeeping on core 0
echo "0" > /sys/fs/cgroup/system.slice/cpuset.cpus
echo "0,1" > /sys/fs/cgroup/user.slice/cpuset.cpus
```

This effectively moves all kernel/system level tasks to only run on core 0, while all user tasks only get scheduled to 0 and 1. However to keep our isolation we still need to use rcu_nocbs=2,3 nohz_full=2,3 and so on. Sadly these seem to not be configurable in Cgroups

Snacj:
fix #1470

### DIFF
--- a/electron/src/helpers/ipc/update/update-listeners.ts
+++ b/electron/src/helpers/ipc/update/update-listeners.ts
@@ -16,7 +16,7 @@ import {
 } from "./update-channels";
 import { spawn, ChildProcess } from "child_process";
 import tkill from "@jub3i/tree-kill";
-import { existsSync, readFileSync, rmSync } from "fs";
+import { existsSync, readFileSync, rmSync, writeFileSync } from "fs";
 import { GithubSource } from "@/setup/GithubSourceDialog";
 import { fetchChangelog, fetchTargets } from "./git-fetch-utils";
 import {
@@ -187,6 +187,10 @@ export function addUpdateEventListeners() {
         });
 
         currentUpdateProcess = null;
+
+        // Re-isolate cores since the build released them
+        await isolateCores(event);
+
         event.sender.send(UPDATE_END, terminalInfo("Update process cancelled"));
         return { success: true };
       } catch (error: any) {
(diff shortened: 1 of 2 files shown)


ninelore (changes requested):
please remove the dependabot and mergify commits and rebase properly

electron/src/helpers/ipc/update/update-listeners.ts
```diff
@@ -1030,3 +1043,119 @@ function terminalInfo(text: string): string {
 function terminalGray(text: string): string {
   return terminalColor("gray", text);
 }
+
+// CPU isolation helpers
+
+const RT_CPUS = "2-3";
+const HK_CPUS = "0-1";
+const ALL_CPUS = "0-3";
```
ninelore: core count shouldnt be hardcoded, this might cause issues when a different core count is present, which likely will happen if we switch to CRA-compliant HMI hardware.

Suggestion: use the `nproc` command and calculate with that

electron/src/helpers/ipc/update/update-listeners.ts
```diff
@@ -1030,3 +1043,119 @@ function terminalInfo(text: string): string {
...
+    `for cpu in /sys/devices/system/cpu/cpu*/cpufreq/scaling_governor; do`,
+    `  echo performance > "$cpu"`,
+    `done`,
+    `for cpu in 2 3; do`,
+    `  max=$(cat /sys/devices/system/cpu/cpu$cpu/cpufreq/cpuinfo_max_freq)`,
+    `  echo $max > /sys/devices/system/cpu/cpu$cpu/cpufreq/scaling_min_freq`,
+    `  echo performance > /sys/devices/system/cpu/cpu$cpu/cpufreq/energy_performance_preference`,
+    `done`,
+    `echo "=== CPU Status ==="`,
+    `for cpu in 0 1 2 3; do`,
```
ninelore:
```suggestion
    `for cpu in $(seq 0 $(nproc)); do`,
```

pls validate bash syntax, this is a unverified scetch

electron/src/helpers/ipc/update/update-listeners.ts
```diff
@@ -1030,3 +1043,119 @@ function terminalInfo(text: string): string {
...
+  });
+}
+
+async function releaseCores(event: Electron.IpcMainInvokeEvent): Promise<void> {
+  event.sender.send(
+    UPDATE_LOG,
+    terminalInfo("Releasing isolated cores for build..."),
+  );
+
+  const script = [
```
ninelore: add as first line
```js
  `#!/usr/bin/env bash`,
```

electron/src/helpers/ipc/update/update-listeners.ts
```diff
@@ -1030,3 +1043,119 @@ function terminalInfo(text: string): string {
...
+    `  echo $max > /sys/devices/system/cpu/cpu$cpu/cpufreq/scaling_min_freq`,
+    `  echo performance > /sys/devices/system/cpu/cpu$cpu/cpufreq/energy_performance_preference`,
+    `done`,
+    `echo "=== CPU Status ==="`,
+    `for cpu in 0 1 2 3; do`,
+    `  gov=$(cat /sys/devices/system/cpu/cpu$cpu/cpufreq/scaling_governor)`,
+    `  freq=$(cat /sys/devices/system/cpu/cpu$cpu/cpufreq/scaling_cur_freq)`,
+    `  echo "CPU$cpu: governor=$gov freq=$freq"`,
+    `done`,
+  ].join("\n");
```
ninelore: the `join()` might be a unecessaryly expensive operation. Maybe try something like

```js
let multilne_str = `
#!/usr/bin/env bash
echo "Foobar"
`;
```

nixos/os/configuration.nix
```diff
@@ -73,7 +75,7 @@ in
         "qitech.cachix.org-1:l7UTb2FOVAEpRTGTUQTBBPis0JhXZGiANPDKBRbO8Vo="
       ];
       experimental-features = "nix-command flakes";
-      cores = 2;
+      cores = 0; # was 2
```
ninelore: Whats the intention here? If u wanna use all cores, remove the line please

nixos/os/configuration.nix
```diff
@@ -174,6 +176,44 @@ in
...
+  # systemd slice for core cgroup-based isolation
+  systemd.slices.qitech = {
+    description = "QiTech realtime slice";
+    sliceConfig = {
+      AllowedCPUs = "2-3";
+      AllowedMemoryNodes = "0";
+    };
+  };
+
+  systemd.services.cpu-isolate = {
```
ninelore: This is redundant to the implementation in the control server. If u insist on this, comments also apply here

nixos/os/configuration.nix
```diff
@@ -197,6 +237,9 @@ in
     package = pkgs.qitechPackages.server;
   };
 
+  # Force the service into the RT slice
+  systemd.services.qitech.serviceConfig.Slice = lib.mkForce "qitech.slice";
```
ninelore: `lib.mkForce` is unnecessary if not an annoyance here. Please remove.

ninelore (changes requested):
Some out of scope problems:
- We need safeguards to not accidentally alter development machines
- CRA-relevant: We should replace `sudo` calls with `run0` or `pkexec` since these provide a graphical password prompt.

Snacj:
@ninelore please see if the fixes are as you intended. Any suggestions for improvements are greatly appreciated.

TheBest6337:
@ninelore please review, its urgent.

electron/src/helpers/ipc/update/update-listeners.ts
```diff
@@ -1030,3 +1043,108 @@ function terminalInfo(text: string): string {
...
+
+  const script = `\
+#!/usr/bin/env bash
+set -euo pipefail
+
+TOTAL=$(nproc)
+LAST=$((TOTAL - 1))
+RT_START=$((TOTAL - 2))
+RT_CPUS="$RT_START-$LAST"
+HK_CPUS="0-$((RT_START - 1))"
```
ninelore, line 1122: This logic may fall apart and cause issues when dealing with heterogeneous CPU core topologies like Intel Alder Lake and newer as well as basically any recent non-x86 SoC.

Please just add above as a comment since fixing this will likely be a pandoras box.

nixos/os/configuration.nix
```diff
@@ -52,9 +52,11 @@ in
     "oops=panic" # Treat kernel oops as panic for auto-recovery
     "usbcore.autosuspend=-1" # Possibly fixes dre disconnect issue?
 
-    "isolcpus=2,3" # Isolate cpus 2 and 3 from scheduler for better latency, 2 runs ethercatthread and 3 runs server control-loop
+    #"isolcpus=2,3" # Isolate cpus 2 and 3 from scheduler for better latency, 2 runs ethercatthread and 3 runs server control-loop
```
ninelore: Please remove instead of just commenting it out. If we need it again we have the git history.

nixos/os/configuration.nix
```diff
@@ -174,6 +175,15 @@ in
...
   };
 
+  # systemd slice for core cgroup-based isolation
+  systemd.slices.qitech = {
+    description = "QiTech realtime slice";
+    sliceConfig = {
+      AllowedCPUs = "2-3";
+      AllowedMemoryNodes = "0";
+    };
+  };
```
ninelore: Hardcoded cores still here. 

It may be enough to just
```suggestion
  systemd.slices.qitech = { };
```

ninelore (changes requested):

Snacj:
@ninelore i'll need to test if the configuration of cores works at runtime when not declared in configuration.nix maybe we can look at it together when you are in office next time.

nixos/os/configuration.nix
```diff
@@ -52,9 +52,10 @@ in
     "oops=panic" # Treat kernel oops as panic for auto-recovery
     "usbcore.autosuspend=-1" # Possibly fixes dre disconnect issue?
 
-    "isolcpus=2,3" # Isolate cpus 2 and 3 from scheduler for better latency, 2 runs ethercatthread and 3 runs server control-loop
+    "isolcpus=managed_irq,2,3" # managed irq only, no domain flag
     "nohz_full=2,3" # In this mode, the periodic scheduler tick is stopped when only one task is running, reducing kernel interruptions on those CPUs.
-    "rcu_nocbs=2,3" # Moves RCU (Read-Copy Update) callback processing away from CPUs 2 and 3.
+    "irqaffinity=0,1" # keep hardware IRQs off RT cores
+    #"rcu_nocbs=2,3" # Moves RCU (Read-Copy Update) callback processing away from CPUs 2 and 3.
```
ninelore: this one too please

TheBest6337:
@ninelore please review

nixos/os/configuration.nix
```diff
@@ -52,9 +52,9 @@ in
     "oops=panic" # Treat kernel oops as panic for auto-recovery
     "usbcore.autosuspend=-1" # Possibly fixes dre disconnect issue?
 
-    "isolcpus=2,3" # Isolate cpus 2 and 3 from scheduler for better latency, 2 runs ethercatthread and 3 runs server control-loop
+    "isolcpus=managed_irq,2,3" # managed irq only, no domain flag
```
ninelore, line 55: i dont think this is right yet

ninelore (changes requested):
Cores are neither isolated nor released at testing.

TheBest6337:
Because Ninelore is in Holiday I would move this feature to the second next release.

PR #1609, opened 2026-08-20
Fix one axis laser

TheBest6337:

- [ ] Tested locally


Oshgnacknak (approved):
I quite certain it took more then just the laser to create changes in 24 files. If this is due to renaming modules, I'd personally prefer a separate pull for just the rename. Makes the history a lot easier to work with

PR #1635, opened 2026-09-02
Jse aquapath v2

kraemr:
small changes


qitech_control/schemas/aqua_path_v1.yaml
TheBest6337: aqua_path?

PR #1647, opened 2026-09-09
Fix standalone laser

TheBest6337:

- [ ] Tested locally


qitech_control/src/persist.rs
```diff
@@ -2,15 +2,65 @@ use std::fs;
...
+    }
+}
+
+pub fn write_settings(settings: &Settings) -> Result<()> {
+    let json = serde_json::to_string_pretty(settings)?;
+    fs::write(get_settings_path(), json).context("Failed to write settings file")?;
+    Ok(())
+}
+
+pub fn resolve_ethercat_enabled() -> bool {
```
Oshgnacknak, line 49: Env vars should take precedence over config values

qitech_control/src/persist.rs
```diff
@@ -2,15 +2,65 @@ use std::fs;
...
+    if let Some(enabled) = read_settings().ethercat_enabled {
+        return enabled;
+    }
+    match std::env::var("ETHERCAT_ENABLED").as_deref() {
+        Ok("false" | "0" | "no") => false,
+        Ok("true" | "1" | "yes") => true,
+        _ => true,
```
Oshgnacknak: why would anything, including errors since the var is not there resolve to `true`?

Oshgnacknak (dismissed):

just-some-entity (approved):

PR #1649, opened 2026-09-09
Improve fetched release sorting, so that rc's are after major releases

### LINKED ISSUE #1642: BUG: Update tab shows RCs before mayor release of an version
^^

TheBest6337:
Dont know why we did it by name before, but that was stupid lol


ninelore:
We are [SemVer](https://semver.org/) compliant. I'd rather fix the sorting to be compliant.

Sorting by date will have the implication of showing an old tag as newer should it ever get modified.

ninelore (dismissed):
See above

Oshgnacknak (approved):

